PR #14, opened 2019-04-01
fix array after removing member - CinemachineTargetGroup

rivFox:
**Before:**
Try to remove 2 from array [0][1][2][3]

oldTargets: [0][1][2][3]
`Array.Copy(oldTargets, m_Targets, 2);`
m_Targets: [0][1][null]
`if (index < m_Targets.Length - 1)` is false because 2 < 3 - 1 => 2 < 2 => false

m_Targets should be replaced by oldTargets or "-1" isn't necessary


glabute:
Thanks for the input, but this issue is already fixed on master

rivFox:
Ah, right. My mistake.

PR #16, opened 2019-05-09
Fixed endless loop problem if using Ignore Tag

aaBaO:


glabute:
Hey thanks for that.  Nice catch!  Much appreciated.

PR #21, opened 2019-09-12
Fix case [1179842]: Change the way asset editability is checked for components

AscendingMan:
Fixed issue: [1179842](https://issuetracker.unity3d.com/issues/vcs-null-reference-when-setting-chinemachinebrain-custom-blend-via-script-when-using-perforce)

Replaced the conditional with a simple call to AssetDatabase.IsOpenForEdit since that's how asset editability should (ideally) be checked.

### DIFF
--- a/Editor/Utility/EmbeddedAssetHelpers.cs
+++ b/Editor/Utility/EmbeddedAssetHelpers.cs
@@ -103,13 +103,13 @@ namespace Cinemachine.Editor
                     }
                     GUI.enabled = true;
                     if(m_Editor.target != null)
-					{
-						if (!canEditAsset && GUILayout.Button("Check out"))
-						{
-							Task task = Provider.Checkout(AssetDatabase.GetAssetPath(m_Editor.target), CheckoutMode.Asset);
-							task.Wait();
-						}
-					}
+                    {
+			if (!canEditAsset && GUILayout.Button("Check out"))
+			{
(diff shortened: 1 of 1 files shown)


Editor/Utility/EmbeddedAssetHelpers.cs
```diff
@@ -114,10 +102,10 @@ UnityEditor.VersionControl.Asset targetAsset
                             OnChanged(property.objectReferenceValue as T);
                     }
                     GUI.enabled = true;
-                    if (isLockedFile && GUILayout.Button("Check out"))
+                    if (!canEditAsset && GUILayout.Button("Check out"))
```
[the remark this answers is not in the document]
AscendingMan: Fair point, I added a null check to avoid doing that.

Editor/Utility/EmbeddedAssetHelpers.cs
```diff
@@ -114,11 +102,14 @@ UnityEditor.VersionControl.Asset targetAsset
                             OnChanged(property.objectReferenceValue as T);
                     }
                     GUI.enabled = true;
-                    if (isLockedFile && GUILayout.Button("Check out"))
-                    {
-                        UnityEditor.VersionControl.Provider.Checkout(
-                            targetAsset, UnityEditor.VersionControl.CheckoutMode.Both);
-                    }
+                    if(m_Editor.target != null)
```
aras-p: Really strange indentation. Tabs vs spaces?

AscendingMan: Correct. Fixed the accident.

PR #26, opened 2020-01-23
Game View update on Editor value change - bug 1213819

gaborkb:
Addressing this ticket: https://fogbugz.unity3d.com/f/cases/1213819

I added 
```
#if UNITY_EDITOR
    UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
#endif
```
where CMBrainEditor knows the editor value has changed.

**May not be a good solution**, but I could not find any problems caused by this fix, and it solved the problem.

I tested these example scenes: FollowCam, Composer and Timeline scene.

I also found an alternative solution instead of `UnityEditorInternal.InternalEditorUtility.RepaintAllViews()`, here:
https://answers.unity.com/questions/1421572/instantaneous-editorguilayoutcolorfield.html
It claims to be faster. 

Alternative solution:
```
System.Reflection.Assembly assembly = typeof(EditorWindow).Assembly;
System.Type type = assembly.GetType( "UnityEditor.GameView" );
var gameview = EditorWindow.GetWindow(type);

gameview.Repaint();
```
This code could be separated, so we don't get the gameview at every change, just once. E.g. singleton or something like that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Bugfixes
 - Bugfix (1213471, 1213434): add null check in editor
 - Bugfix (1213488): no solo for prefab vcams
+- Bugfix (1213819): RepaintAllViews on editor change
 
 
 ## [2.5.0] - 2020-01-15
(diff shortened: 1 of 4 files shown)


glabute:
For fogbugz fixes, please also add a line to release notes mentioning the bugfix.  
Release notes are located in Changelog.md and Extras~/ReleaseNotes.txt

PR #33, opened 2020-02-12
Dev/body before aim

glabute:
Changed the approach.  New fake stage was causing trouble.
Eliminated it and added a hint on the component to let the vcam handle the special ordering if required

### DIFF
--- a/Runtime/Experimental/CinemachineNewVirtualCamera.cs
+++ b/Runtime/Experimental/CinemachineNewVirtualCamera.cs
@@ -197,20 +197,30 @@ namespace Cinemachine
 
             // Apply the component pipeline
             for (CinemachineCore.Stage stage = CinemachineCore.Stage.Body;
-                stage < CinemachineCore.Stage.Finalize; ++stage)
+                stage <= CinemachineCore.Stage.Finalize; ++stage)
             {
                 var c = m_Components[(int)stage];
                 if (c != null)
                     c.PrePipelineMutateCameraState(ref state, deltaTime);
             }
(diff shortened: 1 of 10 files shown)


Runtime/Experimental/CinemachineNewVirtualCamera.cs
```diff
@@ -255,10 +265,10 @@ protected void UpdateComponentCache()
                 }
             }
 #endif
-            if (m_Components != null && m_Components.Length == (int)CinemachineCore.Stage.Finalize)
+            if (m_Components != null && m_Components.Length == (int)CinemachineCore.Stage.Finalize + 1)
```
gaborkb: Why not add BodyAfterAim to CinemachineCore.Stage instead of adding a 1 to Finalize?


```
public enum Stage
{
    /// <summary>Second stage: position the camera in space</summary>
    Body,

    /// <summary>Third stage: orient the camera to point at the target</summary>
    Aim,

    /// <summary>Optional Fourth stage: for procedural placement that depends on aim</summary>
    BodyAfterAim, 

    /// <summary>Final pipeline stage: apply noise (this is done separately, in the
    /// Correction channel of the CameraState)</summary>
    Noise,

    /// <summary>Not a pipeline stage.  This is invoked on all virtual camera
    /// types, after the pipeline is complete</summary>
    Finalize
};
```

A problem with this solution, because it will give empty loop iterations in a lot of places, e.g.:
```
for (CinemachineCore.Stage stage = CinemachineCore.Stage.Body;stage <= CinemachineCore.Stage.Finalize; ++stage)
{ ... }
```

PR #34, opened 2020-02-18
CmFramingTransposer drifting problem 

gaborkb:


glabute:
hmmm... what happens if group is non-empty, but members have zero radius?

gaborkb:
If the group is in 0,0,0, and the group member is also 0,0,0
- List is empty - problem
- List not empty, but only a None (Transform) in there - works
- List not empty, but 0 radius - works 
- List not empty, non-0 radius - works

With my fix they all work.

But, I checked again, with not 0,0,0 start coordinates, and then there are problems in each case even without my fix (on master).
So I am closing this PR until it's fixed.

gaborkb:
The drifting bug:
Case 1: **Solved by adding an error message box when targets are descendants**
   The system is abused. The Target list in CinemachineTargetGroup script contains Targets that are descendants of the gameobject holding CinemachineTargetGroup script. Drift happens, because the target group updates its position depending on the Target list, but by updating its own position, it also updates the world position of its descendants.
 
Case 2: **Solved by checking for empty group**
When CinemachineTargetGroup is added, the List is Empty, if left like that, it causes a drift in FramingTransposer.

In CinemachineFramingTransposer.cs:

```
if (isGroupFraming)
      followTargetPosition = ComputeGroupBounds(group, ref curState);
```
When group has an empty list, then inside ComputeGroupBounds
`b = GetScreenSpaceGroupBoundingBox(group, ref cameraPos, curState.RawOrientation)`
will return a 0 centered Bound (b). Because it calls `group.GetViewSpaceAngularBounds(observer, out minAngles, out maxAngles, out zRange); `which return zRange (0,0), instead of the correct Z value.

I assume correct Z value is in LastBoundsMatrix:
1 0 0 x
0 1 0 y
0 0 1 Z
0 0 0 1

`LastBoundsMatrix = Matrix4x4.TRS(cameraPos, curState.RawOrientation, Vector3.one);`
This causes a drift here:
`groupCenter = LastBoundsMatrix.MultiplyPoint3x4(b.center);`

I solved it by adding `!group.IsEmpty` condition to the conditions:
`bool isGroupFraming = group != null && !group.IsEmpty && m_GroupFramingMode != FramingMode.None;`

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -1,6 +1,7 @@
 using UnityEngine;
 using System;
 using Cinemachine.Utility;
+using UnityEditor;
```
glabute: Why is this line here?  I think you don't need it

Runtime/Components/CinemachineFramingTransposer.cs
```diff
@@ -422,7 +422,7 @@ public override void MutateCameraState(ref CameraState curState, float deltaTime
 
             // Compute group bounds and adjust follow target for group framing
             ICinemachineTargetGroup group = AbstractFollowTargetGroup;
-            bool isGroupFraming = group != null && m_GroupFramingMode != FramingMode.None;
+            bool isGroupFraming = group != null && !group.IsEmpty && m_GroupFramingMode != FramingMode.None;
```
glabute: Is "IsEmpty" enough?  What if you have a group with a single member that has 0 radius?  I think you will get the same drift. Try checking for group.BoundingSphere.radius == 0 instead.

gaborkb: No drifting there.
I checked empty list, list with nothing in it (None), list with 1 item zero, and non zero radius
list with more items 0 and non zero radius.

Runtime/Behaviours/CinemachineTargetGroup.cs
gaborkb: True

Runtime/Components/CinemachineFramingTransposer.cs
gaborkb: The only time I had drifting when the group was empty

Runtime/Components/CinemachineFramingTransposer.cs
```diff
@@ -422,7 +423,7 @@ public override void MutateCameraState(ref CameraState curState, float deltaTime
 
             // Compute group bounds and adjust follow target for group framing
             ICinemachineTargetGroup group = AbstractFollowTargetGroup;
-            bool isGroupFraming = group != null && m_GroupFramingMode != FramingMode.None;
+            bool isGroupFraming = group != null && Math.Abs(group.Sphere.radius) > 1e-5f && m_GroupFramingMode != FramingMode.None;
```
gaborkb: Hardcoded radius check for 1e-5. What do you think?

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -256,9 +256,6 @@ private static BoundingSphere WeightedMemberBounds(Target t, Vector3 avgPos, flo
         /// </summary>
         public void DoUpdate()
         {
-            if (IsEmpty)
```
gaborkb: IsEmpty check removed, because we need to update mAveragePos and BoundingBox. But, when the list is empty, we only need to update once, and not more.
Should I modify the code to update only once?

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -305,17 +302,19 @@ Vector3 CalculateAveragePosition(out float maxWeight)
```
gaborkb: Changed CalculateAverageOrientation. Tested with a bunch of rotations, and the results seemed more reasonable this way. However, I don't know the use cases of the average rotation on the group. Maybe that could help me determine whether this is a better approach or not.

Also, there is this document for averaging quaternions:
http://wiki.unity3d.com/index.php/Averaging_Quaternions_and_Vectors?_ga=2.247459462.30067401.1582556840-1207821455.1572890629

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -302,17 +302,19 @@ Vector3 CalculateAveragePosition(out float maxWeight)
...
-                    Quaternion q = m_Targets[i].target.rotation;
-                    // This is probably bogus
-                    r = new Quaternion(r.x + q.x * w, r.y + q.y * w, r.z + q.z * w, r.w + q.w * w);
+                    float scaledWeight = m_Targets[i].weight / mMaxWeight; // [0, 1]
+                    r *= Quaternion.Slerp(Quaternion.identity, m_Targets[i].target.rotation, scaledWeight);
+                    weightedAverage += scaledWeight;
                 }
             }
+
+            r = Quaternion.Slerp(Quaternion.identity, r, 1.0f / weightedAverage);
```
glabute: potential divide by 0 if weightedAverage == 0

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -305,17 +302,25 @@ Vector3 CalculateAveragePosition(out float maxWeight)
...
-                    // This is probably bogus
-                    r = new Quaternion(r.x + q.x * w, r.y + q.y * w, r.z + q.z * w, r.w + q.w * w);
+                    float scaledWeight = Math.Abs(mMaxWeight) > UnityVectorExtensions.Epsilon 
+                        ? m_Targets[i].weight / mMaxWeight 
+                        : 1.0f;
+                    r *= Quaternion.Slerp(Quaternion.identity, m_Targets[i].target.rotation, scaledWeight);
+                    weightedAverage += scaledWeight;
                 }
             }
+
```
gaborkb: Remember last valid rotation, instead of prefering identity.

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -305,17 +302,23 @@ Vector3 CalculateAveragePosition(out float maxWeight)
...
             Quaternion r = Quaternion.identity;
             for (int i = 0; i < m_Targets.Length; ++i)
             {
                 if (m_Targets[i].target != null)
                 {
-                    float w = m_Targets[i].weight;
-                    Quaternion q = m_Targets[i].target.rotation;
-                    // This is probably bogus
-                    r = new Quaternion(r.x + q.x * w, r.y + q.y * w, r.z + q.z * w, r.w + q.w * w);
+                    float scaledWeight = m_Targets[i].weight / mMaxWeight;
```
gaborkb: If mMaxWeight is 0. Then scaledWeight is NaN.
Slerp handles it without error.

Runtime/Behaviours/CinemachineTargetGroup.cs
```diff
@@ -305,17 +302,26 @@ Vector3 CalculateAveragePosition(out float maxWeight)
...
                 }
             }
+
+            if (weightedAverage > UnityVectorExtensions.Epsilon)
+                r = Quaternion.Slerp(Quaternion.identity, r, 1.0f / weightedAverage);
+            else
+                r = transform.rotation;
+            
             return r.Normalized();
         }
```
gaborkb: Here is an alternative way of doing the same thing as above.
```
{
    float weightedAverage = 0;
    Quaternion r = Quaternion.identity;
    if (mMaxWeight > UnityVectorExtensions.Epsilon)
    {
        for (int i = 0; i < m_Targets.Length; ++i)
        {
            if (m_Targets[i].target != null)
            {
                float scaledWeight = m_Targets[i].weight / mMaxWeight;
                r *= Quaternion.Slerp(Quaternion.identity, m_Targets[i].target.rotation, scaledWeight);
                weightedAverage += scaledWeight;
            }
        }
        
        r = Quaternion.Slerp(Quaternion.identity, r, 1.0f / weightedAverage);
    }
    else
    {
        r = transform.rotation;
    }

    return r.Normalized();
}
```

gaborkb:
```
{
    float weightedAverage = 0;
    Quaternion r = Quaternion.identity;
    if (mMaxWeight > UnityVectorExtensions.Epsilon)
    {
        for (int i = 0; i < m_Targets.Length; ++i)
        {
            if (m_Targets[i].target != null)
            {
                float scaledWeight = m_Targets[i].weight / mMaxWeight;
                r *= Quaternion.Slerp(Quaternion.identity, m_Targets[i].target.rotation, scaledWeight);
                weightedAverage += scaledWeight;
            }
        }
        
        r = Quaternion.Slerp(Quaternion.identity, r, 1.0f / weightedAverage);
    }
    else
    {
        r = transform.rotation;
    }

    return r.Normalized();
}
```

PR #37, opened 2020-02-20
Blend editor UX improvements: allow direct editing of vcam names

glabute:


glabute:
as per forum request
https://forum.unity.com/threads/display-all-contents-of-cinemachine-blends-asset-even-for-missing-cameras.830980/

gaborkb:
I tried the clear shot demo. Added a custom blend. But I could not add vcams, only the main clear shot vcam. I sent you an image on slack showing what I mean.

PR #44, opened 2020-03-02
SimpleFollowWithWorldUp should not be affected by axis range

gaborkb:
https://fogbugz.unity3d.com/f/cases/1222740/

Currently, BindingMode.SimpleFollowWithWorldUp is affected by axis's range (min, max), which causes undesired behaviour. 
The axis range parameters are applied in `CinemachineOrbitalTransposer.cs`, when `axis.Update(deltaTime)` is called.

`axis.Update(deltaTime)` calls `bool MaxSpeedUpdate(float input, float deltaTime)` in `AxisState.cs`. `MaxSpeedUpdate` updates the input axis value based on the min, max value range.


glabute:
What about a vcam with an orbital transposer? same thing will happen

PR #45, opened 2020-03-02
Confiner cache reset

gaborkb:
When BoundingShape2D/BoundingVolume is changed, we need to invalidate the cache.

I created a getter/setter for these parameters. When they are set, the cache is invalidated.

However, when the property is changed in the editor, it is not changed through the setter, so we need to invalidate cache in the OnValidate method.

QUESTION:
Do we want to expose `InvalidatePathCache` function after this change?
```
/// <summary>Call this if the bounding shape's points change at runtime</summary>
public void InvalidatePathCache() { m_pathCache = null; }
```


glabute:
I think a better fix would be to cache the source collider when the path cache is created.  Cache is valid if the cached collider matches the current collider.  Then you don't need an API change.

We still need to keep InvalidatePathCache() because the user might modify the points of the collider at runtime, and we can't detect that

glabute:
why did you close this?

gaborkb:
While I worked on it, I closed it, but now it's open again

PR #49, opened 2020-04-01
Add ForceCameraPosition method to vcams

glabute:
This is an API to allow clients to manually initialize vcam position and orientation.  This has been requested multiple times on the forums.


glabute:
Re-implemented more thoroughly - should work now

PR #51, opened 2020-04-01
Confiner jerk bug

gaborkb:
Tested and seems like in both 2D and 3D modes, the jerk bug is there if we allow stages before Finalize.

What do you think?

### DIFF
--- a/Runtime/Behaviours/CinemachineConfiner.cs
+++ b/Runtime/Behaviours/CinemachineConfiner.cs
@@ -82,6 +82,7 @@ namespace Cinemachine
         {
             public Vector3 m_previousDisplacement;
             public float confinerDisplacement;
+            public bool isFramingTransposer;
         };
 
         /// <summary>Check if the bounding volume is defined</summary>
@@ -100,6 +101,24 @@ namespace Cinemachine
             }
         }
 
+        protected override void ConnectToVcam(bool connect)
+        {
+            base.ConnectToVcam(connect);
+
+            CinemachineVirtualCamera vcam = VirtualCamera as CinemachineVirtualCamera;
+            var components = vcam.GetComponentPipeline();
+            foreach (var component in components)
+            {
+                if (component.BodyAppliesAfterAim)
+                {
+                    var extraState = GetExtraState<VcamExtraState>(VirtualCamera);
+                    extraState.isFramingTransposer = true;
+                    break;
+                }
+            }
+        }
+        
+
         /// <summary>Callback to to the camera confining</summary>
         protected override void PostPipelineStageCallback(
             CinemachineVirtualCameraBase vcam,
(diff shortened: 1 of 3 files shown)


gaborkb:
Please check with transposer too. I checked on master vs this branch using Transposer, and both branches produce same output; but they are both strange.

Framing Transposer does not have the jerk problem anymore.

Runtime/Behaviours/CinemachineConfiner.cs
```diff
@@ -82,6 +82,7 @@ class VcamExtraState
         {
             public Vector3 m_previousDisplacement;
             public float confinerDisplacement;
+            public bool isFramingTransposer;
```
glabute, line 85: I would call this applyAfterAim - it's not necessarily the framing transposer, there could be another component with that property

Runtime/Behaviours/CinemachineConfiner.cs
```diff
@@ -100,15 +101,36 @@ public bool IsValid
             }
         }
 
+        protected override void ConnectToVcam(bool connect)
+        {
+            base.ConnectToVcam(connect);
+
+            CinemachineVirtualCamera vcam = VirtualCamera as CinemachineVirtualCamera;
+            var components = vcam.GetComponentPipeline();
```
glabute, line 109: vcam can be null, you have to check if the cast succeeded

Runtime/Behaviours/CinemachineConfiner.cs
```diff
@@ -100,15 +101,36 @@ public bool IsValid
...
         protected override void PostPipelineStageCallback(
             CinemachineVirtualCameraBase vcam,
             CinemachineCore.Stage stage, ref CameraState state, float deltaTime)
         {
             if (IsValid)
             {
-                // Move the body before the Aim is calculated
-                if (stage == CinemachineCore.Stage.Body)
+                var extraState = GetExtraState<VcamExtraState>(VirtualCamera);
+                if (extraState.isFramingTransposer && stage == CinemachineCore.Stage.Finalize
```
glabute, line 130: do you need brackets here?

Runtime/Behaviours/CinemachineConfiner.cs
```diff
@@ -100,15 +101,36 @@ public bool IsValid
...
         /// <summary>Callback to to the camera confining</summary>
         protected override void PostPipelineStageCallback(
             CinemachineVirtualCameraBase vcam,
             CinemachineCore.Stage stage, ref CameraState state, float deltaTime)
         {
             if (IsValid)
             {
-                // Move the body before the Aim is calculated
-                if (stage == CinemachineCore.Stage.Body)
+                var extraState = GetExtraState<VcamExtraState>(VirtualCamera);
```
glabute, line 129: you have to use vcam here, not VirtualCamera.  See the correct call a few lines below.  Also, why get extraState twice?

Runtime/Behaviours/CinemachineConfiner.cs
glabute: also: you need to use vcam, not VirtualCamera as parameter, to get the right extraState.
It's done just below - you should use that, no need to get extraState twice.

glabute: ok, you don't *need* brackets, but you should put them, because nobody remembers the order of evaluation of && and || (at least I don't :) )

gaborkb: No brackets are needed, however I agree it is more clear if I put brackets.

Runtime/Behaviours/CinemachineConfiner.cs
gaborkb: Totally true!

gaborkb:
Have not added change log yet!

PR #60, opened 2020-04-21
incorrect sanity check

whoisfpc:
Because `enable = false` will call `OnDisable` function immediately, we need move sanity check statements to below normal enable statements.


glabute:
Nice catch.  Thank you for that.

PR #62, opened 2020-04-28
Dev/new input system

glabute:
Create a new hook for overriding the input system.  AxisState will call this, if present, instead of CinemachineCore.GetInputAxis().  It allows for a more generalized support of input systems, because the implementer can have extra state in their class.

An example implementation is CinemachineInputProvider, which implements for the UniteEngine.Input package

Feature description: https://docs.google.com/document/d/1yRLdDp0OVAQPQ2jjDwwxNg9nIkE13KokwnEDETSyzoo/edit#

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -17,3 +17,4 @@ upm-ci~/**
 *.zip*
 TestRunnerOptions.json
 .idea/**
+upm-ci.log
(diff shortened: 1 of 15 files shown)


glabute:
I think QA will do those tests (i.e. Patrick).  You just have to do a code review and basic testing.

PR #65, opened 2020-05-12
Dev/timeline scrubbing

glabute:
Added optional timeline scrubbing cache feature.  
Description is here:
https://docs.google.com/document/d/1wb2V-VXPkgsWNgXb6jCF8wKtvGWY66JWaMjAC4y77ww/edit#heading=h.jxyw25p59zfr

### DIFF
--- a/Runtime/Core/TargetPositionCache.cs
+++ b/Runtime/Core/TargetPositionCache.cs
@@ -47,42 +47,53 @@ namespace Cinemachine
             }
             List<RecordingItem> RawItems = new List<RecordingItem>();
 
+            const float kResolution = 0.04f;
+
             public void AddRawItem(float time, Transform target)
             {
-                RawItems.Add(new RecordingItem
-                {
-                    Time = time,
-                    Pos = target.position,
-                    Rot = target.rotation.eulerAngles
-                });
+                var n = RawItems.Count;
+                if (n == 0 || Mathf.Abs(RawItems[n-1].Time - time) >= kResolution)
+                    RawItems.Add(new RecordingItem
+                    {
+                        Time = time,
+                        Pos = target.position,
+                        Rot = target.rotation.eulerAngles
+                    });
             }
 
             public void CreateCurves()
             {
-                X = new AnimationCurve();
-                Y = new AnimationCurve();
-                Z = new AnimationCurve();
-                RotX = new AnimationCurve();
-                RotY = new AnimationCurve();
-                RotZ = new AnimationCurve();
+                var XList = new List<Keyframe>();
+                var YList = new List<Keyframe>();
+                var ZList = new List<Keyframe>();
+                var RotXList = new List<Keyframe>();
+                var RotYList = new List<Keyframe>();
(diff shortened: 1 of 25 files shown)


Runtime/Core/TargetPositionCache.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.Collections.Generic;
+using System;
+
+namespace Cinemachine
+{
+    internal class TargetPositionCache
+    {
+        public static bool UseCache { get; set; }
+        public enum Mode { Disabled, Record, Playback }
+        public static Mode m_CacheMode = Mode.Disabled;
```
gaborkb, line 11: m_CacheMode could be private

Runtime/Core/TargetPositionCache.cs
```diff
@@ -0,0 +1,244 @@
...
+                var ZList = new List<Keyframe>();
+                var RotXList = new List<Keyframe>();
+                var RotYList = new List<Keyframe>();
+                var RotZList = new List<Keyframe>();
+
+                RawItems.Sort();
+                float time = float.MaxValue;
+                for (int i = 0; i < RawItems.Count; ++i)
+                {
+                    var item = RawItems[i];
```
gaborkb, line 77: foreach (var item in RawItems)
{

Foreach may be nicer, but it's not necessarily better.

Runtime/Timeline/CinemachineMixer.cs
```diff
@@ -6,26 +6,139 @@
...
+                    {
+                        var vcam = sublist[j];
+                        maxDampingTime = Mathf.Max(maxDampingTime, vcam.GetMaxDampTime());
+                    }
+                }
+                // Impose upper limit on damping time, to avoid simulating too many frames
+                return Mathf.Min(maxDampingTime, 2.0f); 
+            }
+
+            public void ScrubTohere(
```
gaborkb, line 65: ScrubToHere - casing typo

Runtime/Core/TargetPositionCache.cs
glabute: yes

Runtime/Core/TargetPositionCache.cs
glabute: I used to use foreach, but then discovered that there are situations where it does GC allocs.  So now I always do it this way

Runtime/Timeline/CinemachineMixer.cs
glabute: yes

glabute:
I'm worrying a little about performance.  I think AnimationCurve is very inefficient when it evaluates (it has to iterate all keys from beginning to find the desired time, every frame).  Maybe a custom implementation of a segmented linear with a fixed timestep would be better

Runtime/Core/TargetPositionCache.cs
gaborkb: Thanks for the info! :)

PR #66, opened 2020-05-14
Remove AnimationCurves, 

glabute:
replace with custom implementation optimized for fixed step size

### DIFF
--- a/Runtime/Core/TargetPositionCache.cs
+++ b/Runtime/Core/TargetPositionCache.cs
@@ -30,20 +30,78 @@ namespace Cinemachine
 
         public static float CurrentTime { get; set; }
 
+        class CacheCurve
+        {
+            public struct Item
+            {
+                public Vector3 Pos;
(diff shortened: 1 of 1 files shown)


Runtime/Core/TargetPositionCache.cs
```diff
@@ -30,20 +30,78 @@ public static Mode CacheMode
...
+                var index = Mathf.Max(Mathf.FloorToInt(s / StepSize), 0);
+                if (index >= numItems - 1)
+                    return m_Cache[numItems - 1];
+
+                float t = (s - (index * StepSize)) / StepSize;
+                return Item.Lerp(m_Cache[index], m_Cache[index + 1], t);
+            }
+        }
+
         class CacheEntry
```
gaborkb, line 97: CacheEntry class could be private.

glabute: it is private by default

PR #67, opened 2020-05-15
Remove some GC allocs when in edit mode (reported by customer)

glabute:
Reported by customer.
Fixed for 2019.2 and up, using new TryGetComponent API

### DIFF
--- a/Runtime/PostProcessing/CinemachinePostProcessing.cs
+++ b/Runtime/PostProcessing/CinemachinePostProcessing.cs
@@ -276,8 +276,8 @@ namespace Cinemachine.PostFX
 
         static PostProcessLayer GetPPLayer(CinemachineBrain brain)
         {
-            PostProcessLayer layer = null;
-            if (mBrainToLayer.TryGetValue(brain, out layer))
+            bool found = mBrainToLayer.TryGetValue(brain, out PostProcessLayer layer);
+            if (found)
             {
 #if UNITY_EDITOR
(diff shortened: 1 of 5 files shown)


Runtime/PostProcessing/CinemachinePostProcessing.cs
```diff
@@ -276,21 +276,30 @@ static void ApplyPostFX(CinemachineBrain brain)
...
+#if UNITY_2019_2_OR_NEWER
+            brain.TryGetComponent(out layer);
+#else
             layer = brain.GetComponent<PostProcessLayer>();
-            mBrainToLayer[brain] = layer;
-            if (layer != null)
-                brain.m_CameraCutEvent.AddListener(OnCameraCut);
-            else
+#endif
+            if (found)
```
gaborkb, line 293: if found is true, then we returned at line 286.
Except, if UNITY_EDITOR is true, and (layer != null || Application.isPlaying) is false.

Is this intentional?

glabute: Yes.  If not playing, we allow for the possibility of the user adding a PP layer at any time, because they are still editing the scene.  We don't bother with this check if playing, to save on processing time where it matters.

glabute: I agree that this code is very confusing... and not quite correct.  I will rewrite

PR #69, opened 2020-06-12
Dev/confiner corner damping

gaborkb:


Runtime/Behaviours/CinemachineConfiner.cs
```diff
@@ -155,13 +165,28 @@ public override float GetMaxDampTime()
...
+                    
+                   
+                    if (VirtualCamera.PreviousStateIsValid && deltaTime >= 0)
+                    { 
+                        var displacementAngle = Vector2.Angle(extra.m_previousDisplacement, displacement);
+                        if (m_CornerDamping > 0 && (m_Cornerring || displacementAngle > m_CornerAngleTreshold))
+                        {
+                            if (!m_SnapFromCorner) {
+                                m_Cornerring = displacementAngle > 1f;
+                            }
```
gaborkb, line 177: I am not sure if the option to turn this on/off is needed.
I think it should always be the smoother option, but I could not decide.

glabute: Can you tell me what this PR is about?  What specific issue is it addressing?

gaborkb: Users (e.g. https://forum.unity.com/threads/confine-screen-edges-weird-bounding-shapes-bad-results.902894/#post-5925617) want to have damping at inside corners, so the camera does not snap.

To decide when to do corner damping, I check the angle between the previous displacement vector and the current displacement vector (extra.m_previousDisplacement, displacement).
The treshold for the angle is 10 degrees stored in m_CornerAngleTreshold. I just found it to be a good number in my test scenes.

PR #71, opened 2020-06-18
Replaced direct call to AnimationCurve.Linear

gaborkb:
https://fogbugz.unity3d.com/f/cases/1252431/

To fix this, I replaced the call to AnimationCurve.Linear every frame to a static variable.


glabute:
Please add release notes (new CM version: 2.6.1)

PR #74, opened 2020-07-02
Bugfix: StateDrivenCamera was choosing parent state if only 1 clip in…

glabute:
… blendstate


glabute:
It's addressing this issue:
https://forum.unity.com/threads/camera-not-being-set-on-state-driven-camera-and-blend-tree.646999/#post-6016676

The check was wrong: it prevented the substate from being recognized in cases where there is only 1 substate

PR #79, opened 2020-07-16
[Editor] Enable multi-editing of Cinemachine components/extensions, where applicable

AscendingMan:
-----
### Purpose of this PR

Added multi-selection support to Cinemachine editors where it made sense, including:

- Virtual Camera
- Free Look Camera
- Blend List, Clear Shot Cameras (support is limited due to these pointing to child objects of the GO to which the camera component is attached, thus there is no multi-editing for that)
- All of the Extensions
------
### Testing status

Tested manually to work when handling multiple objects by myself and our internal team QA.

### Screenshots
CinemachineMultiEdti

### DIFF
--- a/Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
+++ b/Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
@@ -3,6 +3,7 @@ using UnityEditor;
 using System;
 using System.Collections.Generic;
 using Cinemachine.Utility;
+using UnityEngine.UIElements;
 
 namespace Cinemachine.Editor
 {
@@ -13,6 +14,12 @@ namespace Cinemachine.Editor
     public class CinemachineVirtualCameraBaseEditor<T>
         : BaseEditor<T> where T : CinemachineVirtualCameraBase
     {
+        public static class Styles
+        {
+            public static GUIContent addExtensionLabel = new GUIContent("Add Extension");
+            public static GUIContent virtualCameraChildrenInfoMsg = new GUIContent("The Virtual Camera Children field displays the child objects of a given parent object, thus this field is not available when multiple objects are selected.");
+        }
+        
         static Type[] sExtensionTypes;  // First entry is null
         static string[] sExtensionNames;
         bool IsPrefabBase { get; set; }
@@ -111,14 +118,27 @@ namespace Cinemachine.Editor
                 EditorGUILayout.Space();
                 EditorGUILayout.LabelField("Extensions", EditorStyles.boldLabel);
                 Rect rect = EditorGUILayout.GetControlRect(true, EditorGUIUtility.singleLineHeight);
-                rect = EditorGUI.PrefixLabel(rect, new GUIContent("Add Extension"));
+                rect = EditorGUI.PrefixLabel(rect, Styles.addExtensionLabel);
 
                 int selection = EditorGUI.Popup(rect, 0, sExtensionNames);
                 if (selection > 0)
                 {
                     Type extType = sExtensionTypes[selection];
-                    if (Target.GetComponent(extType) == null)
-                        Undo.AddComponent(Target.gameObject, extType);
+
+                    if (Selection.objects.Length == 1)
+                    {
+                        if (Target.GetComponent(extType) == null)
+                            Undo.AddComponent(Target.gameObject, extType);
+                    }
+                    else
+                    {
+                        for (int i = 0; i < targets.Length; i++)
+                        {
+                            var targetGO = (targets[i] as CinemachineVirtualCameraBase).gameObject; 
+                            if (targetGO != null && targetGO.GetComponent(extType) == null)
+                                Undo.AddComponent(targetGO, extType);
+                        }
+                    }
                 }
                 ExcludeProperty("Extensions");
             }
@@ -126,6 +146,9 @@ namespace Cinemachine.Editor
 
         protected void DrawCameraStatusInInspector()
         {
+            if (Selection.objects.Length > 1)
+                return;
+            
             // Is the camera navel-gazing?
             CameraState state = Target.State;
             if (state.HasLookAt && (state.ReferenceLookAt - state.CorrectedPosition).AlmostZero())
--- a/Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
+++ b/Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
@@ -3,7 +3,6 @@ using UnityEditor;
 using System;
 using System.Collections.Generic;
 using Cinemachine.Utility;
-using UnityEngine.UIElements;
 
 namespace Cinemachine.Editor
 {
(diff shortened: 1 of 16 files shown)


AscendingMan:
Cc @glabute @aras-p For exposure as I can't add reviewers for some reason >_<

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using Cinemachine.Utility;
+using UnityEngine.UIElements;
```
aras-p, line 6: Minor: why using of UIElements namespace is needed here?

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
```diff
@@ -111,21 +118,37 @@ protected void DrawExtensionsWidgetInInspector()
...
+                rect = EditorGUI.PrefixLabel(rect, Styles.addExtensionLabel);
 
                 int selection = EditorGUI.Popup(rect, 0, sExtensionNames);
                 if (selection > 0)
                 {
                     Type extType = sExtensionTypes[selection];
-                    if (Target.GetComponent(extType) == null)
-                        Undo.AddComponent(Target.gameObject, extType);
+
+                    if (Selection.objects.Length == 1)
```
aras-p, line 128: Why this needs separate code paths for 1 vs non-1 length? From a quick look it feels like the "loop over targets" would just work fine for 1 length as well?

Editor/Editors/CinemachineVirtualCameraEditor.cs
```diff
@@ -40,6 +42,14 @@ protected override void OnEnable()
             base.OnEnable();
             IsPrefab = Target.gameObject.scene.name == null; // causes a small GC alloc
             UpdateStaticData();
+            UpdateStageDataTypeMatchesForMultiSelection();
+            Undo.undoRedoPerformed += ResetTargetOnUndo;
```
aras-p, line 46: Does this ever de-register undoRedoPerformed handler?

Editor/Editors/CinemachineVirtualCameraEditor.cs
```diff
@@ -390,9 +443,21 @@ void UpdateComponentEditors()
...
+                        if (cinemachineVirtualCamera == null)
+                            continue;
+
+                        var behaviour = cinemachineVirtualCamera.GetCinemachineComponent(components[i].Stage) as MonoBehaviour;
+                        if (behaviour != null)
+                            behaviours.Add(behaviour);
+                    }
+
+                    var behaviourArray = behaviours.ToArray();
+                    if (behaviourArray.Length > 0 && behaviourArray[0] != null && (hasSameStageDataTypes[(int)components[i].Stage] || targets.Length == 1))
```
aras-p, line 459: Why the `behaviourArray[0] != null` check is needed? I don't quite see how would a null entry get in there

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
AscendingMan: Rider accident.

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
AscendingMan: I missed that when refactoring, don't think there's any other reason to it.

Editor/Editors/CinemachineVirtualCameraEditor.cs
AscendingMan: Yep, leftovers from when I used to cast it to a different type. Will fix.

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
```diff
@@ -13,6 +14,12 @@ namespace Cinemachine.Editor
     public class CinemachineVirtualCameraBaseEditor<T>
         : BaseEditor<T> where T : CinemachineVirtualCameraBase
     {
+        public static class Styles
+        {
+            public static GUIContent addExtensionLabel = new GUIContent("Add Extension");
+            public static GUIContent virtualCameraChildrenInfoMsg = new GUIContent("The Virtual Camera Children field displays the child objects of a given parent object, thus this field is not available when multiple objects are selected.");
```
AscendingMan, line 20: This text will need to be fixed not to display "Virtual Camera" for cases when it's on a different type of camera.

Editor/Editors/CinemachineVirtualCameraEditor.cs
```diff
@@ -153,16 +163,28 @@ protected void DrawPipelineInInspector()
...
                 r = rect; r.width -= labelWidth; r.x += labelWidth;
                 GUI.enabled = !StageIsLocked(stage);
-                int newSelection = EditorGUI.Popup(r, m_stageState[index], sStageData[index].PopupOptions);
+
+                if (targets.Length > 1)
+                    EditorGUI.showMixedValue = !hasSameStageDataTypes[index];
+
                 GUI.enabled = true;
+                EditorGUI.BeginChangeCheck();
+                int newSelection = EditorGUI.Popup(r, m_stageState[index], sStageData[index].PopupOptions);
```
glabute, line 172: GUI.enabled = !StageIsLocked(stage) above is supposed to apply to the popup - you have broken that by setting GUI.enabled = true.

FreeLook, for example, does not allow the user to change the popup value for Body

image

Editor/Editors/CinemachineVirtualCameraEditor.cs
```diff
@@ -382,25 +378,57 @@ bool useSimple
             }
         }
 
+        List<CinemachineComponentBase> GetSortedPipelineComponentsForCamera(UnityEngine.Object targetObject)
+        {
+            var camera = targetObject as CinemachineVirtualCamera;
+            if (camera == null)
+                return null;
+
+            var cameraComponents = new List<CinemachineComponentBase>();
```
glabute, line 387: This code is creating a lot of GC allocs, which are very bad for editor performance.

glabute: You can replace the whole thing with this, which is much more efficient:

```
        void GetPipelineTypes(CinemachineVirtualCamera vcam, ref Type[] types)
        {
            for (int i = 0; i < types.Length; ++i)
                types[i] = null;
            if (vcam != null)
            {
                var components = vcam.GetComponentPipeline();
                for (int j = 0; j < components.Length; ++j)
                    types[(int)components[j].Stage] = components[j].GetType();
            }
        }

        // scratch buffers for pipeline types
        Type[] s_PipelineTypeCache0 = new Type[Enum.GetValues(typeof(CinemachineCore.Stage)).Length];
        Type[] s_PipelineTypeCacheN = new Type[Enum.GetValues(typeof(CinemachineCore.Stage)).Length];

        void UpdateStageDataTypeMatchesForMultiSelection()
        {
            for (int i = 0; i < m_hasSameStageDataTypes.Length; ++i)
                m_hasSameStageDataTypes[i] = true;

            if (targets.Length > 1)
            {
                GetPipelineTypes(serializedObject.targetObjects[0] as CinemachineVirtualCamera, ref s_PipelineTypeCache0);
                for (int i = 1; i < targets.Length; ++i)
                {
                    GetPipelineTypes(serializedObject.targetObjects[i] as CinemachineVirtualCamera, ref s_PipelineTypeCacheN);
                    for (int j = 0; j < s_PipelineTypeCache0.Length; ++j)
                        if (s_PipelineTypeCache0[j] != s_PipelineTypeCacheN[j])
                            m_hasSameStageDataTypes[j] = false;
                }
            }
        }
```

glabute:
Interesting.  I am also on VS 2019, and I get this warning.  Maybe you should check if your VS setup is correct?

aras-p:
Would this need a changelog entry btw?

glabute:
yeah, I'll take care of it

PR #85, opened 2020-07-30
Optimize path gizmos

aras-p:
- Only render full "railroad" track for currently active selection,
- Small micro-optimizations in path evaluation code.

On BoatAttack project, cuts down CM path gizmo rendering time (per camera - there's two cameras in that project) from 4.06ms down to 0.79ms on my machine.

### DIFF
--- a/Runtime/Core/SplineHelpers.cs
+++ b/Runtime/Core/SplineHelpers.cs
@@ -17,7 +17,7 @@ namespace Cinemachine.Utility
             float t, Vector3 p0, Vector3 p1, Vector3 p2, Vector3 p3)
         {
             t = Mathf.Clamp01(t);
-            return (-3f * p0 + 9f * p1 - 9f * p2 + 3f * p3) * t * t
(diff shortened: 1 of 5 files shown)


Runtime/Core/SplineHelpers.cs
```diff
@@ -17,7 +17,7 @@ public static class SplineHelpers
             float t, Vector3 p0, Vector3 p1, Vector3 p2, Vector3 p3)
         {
             t = Mathf.Clamp01(t);
-            return (-3f * p0 + 9f * p1 - 9f * p2 + 3f * p3) * t * t
+            return (-3f * p0 + 9f * p1 - 9f * p2 + 3f * p3) * (t * t)
```
lucasmeijer, line 20: why is this faster?

aras-p: `((vector * float) * float)` vs `vector * (float * float)` -- this was actually pointed out by Rider and I auto-accepted it

PR #87, opened 2020-08-02
Update cinemachineAPI.md

midopooler:
Fixed 27 typos


shivaylamba:
@midopooler nice work, you should be added to the org

PR #88, opened 2020-08-04
Regression fix: Confiner no longer confines noise and impulse

glabute:
Fix for CMCL-107, reported here:
https://forum.unity.com/threads/confiner-impulse-listener-dont-play-nice-together.940029/#post-6159074

### DIFF
--- a/Runtime/Behaviours/CinemachineVirtualCamera.cs
+++ b/Runtime/Behaviours/CinemachineVirtualCamera.cs
@@ -65,7 +65,8 @@ namespace Cinemachine
         /// will refer to this target and orient the vcam in accordance with rules and
         /// settings that are provided to it.
         /// If this is null, then the vcam's Transform orientation will be used.</summary>
-        [Tooltip("The object that the camera wants to look at (the Aim target).  If this is null, then the vcam's Transform orientation will define the camera's orientation.")]
+        [Tooltip("The object that the camera wants to look at (the Aim target).  "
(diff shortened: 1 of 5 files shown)


Runtime/Behaviours/CinemachineVirtualCamera.cs
```diff
@@ -493,27 +493,34 @@ private CameraState CalculateNewState(Vector3 worldUp, float deltaTime)
...
+                    if (stage == CinemachineCore.Stage.Aim)
+                    {
+                        if (c == null)
+                            state.BlendHint |= CameraState.BlendHintValue.IgnoreLookAtTarget;
+                        if (postAimBody != null)
+                        {
+                            postAimBody.MutateCameraState(ref state, deltaTime);
+                            InvokePostPipelineStageCallback(this, CinemachineCore.Stage.Body, ref state, deltaTime);
+                        }
+                    }
```
gaborkb, line 523: This part is a bit difficult to understand at first look. It looks like _InvokePostPipelineStageCallback_ is called twice if stage is CinemachineCore.Stage.Aim. But postAimBody is null in that case, so it's ok. :)

glabute: What is supposed to happen is that if there is a Body component with the BodyAppliesAfterAim property, the pipeline order is reversed: Aim then Body instead of Body then Aim.  Does that make sense?

PR #95, opened 2020-08-14
Collider must be enabled for the confiner

gaborkb:


gaborkb:
.enabled does not return false, when the gameObject is deactivated, but the component is enabled. Good point! I added for activeInHierarchy, which return false if the gameobject or its parents are deactivated.

glabute (approved):

PR #98, opened 2020-08-26
pipeline can be null, in that case we get an error in console

gaborkb:
Bug reported: https://fogbugz.unity3d.com/f/cases/1272146/


glabute:
please add release note in master

PR #102, opened 2020-09-15
Bugfix: BlendList camera was incorrectly holding 0-length camera cuts

glabute:
Bugfix for issue reported here:
https://forum.unity.com/threads/unexpected-delay-in-blend-list-camera-cuts.970029/#post-6312258


glabute:
If you're wondering why I changed more than CinemachineBlendList.cs, the answer is that I did a search for m_Time and fixed all the places where it was being read

gaborkb (approved):

PR #104, opened 2020-09-22
Add Unity.Spline Support in Cinemachine

jieyingLin:


Editor/Menus/CinemachineMenu.cs
```diff
@@ -144,7 +145,7 @@ private static void CreateDollyTrackWithCart()
             if (SceneView.lastActiveSceneView != null)
                 go.transform.position = SceneView.lastActiveSceneView.pivot;
             Undo.RegisterCreatedObjectUndo(go, "create track");
-            CinemachineSmoothPath path = go.GetComponent<CinemachineSmoothPath>();
+            SplineContainer path = go.GetComponent<SplineContainer>();
```
glabute, line 148: This will return null because you haven't added the component.  This needs to be done a few lines above, instead of creating the CinemachineSmoothPath component

glabute: Did you try running this code by selecting the Create Dolly Cart item in the Cinemachine menu?

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -20,7 +22,10 @@ public class CinemachineDollyCart : MonoBehaviour
     {
         /// <summary>The path to follow</summary>
         [Tooltip("The path to follow")]
-        public CinemachinePathBase m_Path;
+        public SplineContainer m_Path;
+        //public CinemachinePathBase m_Path;
```
glabute, line 26: you can't leave in commented-out lines.  I assume that if this were a real PR, these wouldn't be here

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -74,12 +79,33 @@ void LateUpdate()
...
+                //transform.position = m_Path.EvaluatePositionAtUnit(m_Position, m_PositionUnits);
+                //transform.rotation = m_Path.EvaluateOrientationAtUnit(m_Position, m_PositionUnits);
+            //}
             if (m_Path != null)
             {
-                m_Position = m_Path.StandardizeUnit(distanceAlongPath, m_PositionUnits);
-                transform.position = m_Path.EvaluatePositionAtUnit(m_Position, m_PositionUnits);
-                transform.rotation = m_Path.EvaluateOrientationAtUnit(m_Position, m_PositionUnits);
+                m_Position = StandardizePathDistance(distanceAlongPath);
+                float3 result = m_Path.EvaluatePosition(m_Position / m_Path.CalculateLength());
```
glabute, line 91: you're calculating the path length twice - once in Standardize, and once here.  Not efficient

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -74,12 +79,33 @@ void LateUpdate()
...
+                transform.position = new Vector3(result.x, result.y, result.z);
+                //transform.rotation = m_Path.EvaluateOrientationAtUnit(m_Position, m_PositionUnits);
+            }
+        }
+
+        float StandardizePathDistance(float distance)
```
glabute, line 97: as discussed, this should be moved to a helper class, because TrackedDolly will need it too

Runtime/com.unity.cinemachine.asmdef
```diff
@@ -1,14 +1,17 @@
 {
     "name": "Cinemachine",
+    "rootNamespace": "",
     "references": [
-        "Unity.Timeline",
+        "GUID:f06555f75b070af458a003d92f9efb00",
```
glabute, line 5: Somehow you had "use GUIDS" checked in the asmdef inspector.  We want to use human-readable names, not GUIDS

glabute (changes requested):

Editor/Menus/CinemachineMenu.cs
jieyingLin: My bad, should be fixed in next commit.

Runtime/Behaviours/CinemachineDollyCart.cs
jieyingLin: You are right. I come up with two ways to solve this problem:
1. Pass length as a valuable to both functions
2. Merge `StandardizeUnit` and `EvaluatePositionAtUnit` into one function

Which one do you think is better?

glabute: Either approach has its advantages, so the decision will depend on the surrounding code.  I suggest that, for ease of conversion from the old system, you build a helper class with an API as close as possible to CinemachinePathBase.  That way, converting legacy code (such as DollyCart) will be trivial.  Don't forget that our customers will have to do something similar if they have custom classes that use paths.

jieyingLin: What about keep both `StandardizeUnit` and `EvaluatePositionAtUnit` in the helper class? But in `EvaluatePositionAtUnit`, add `StandardizeUnit` before it (copy the code, not calling the function)
Advantages:
1. customers don't need to remember to call `StandardizeUnit` every time they want to use `EvaluatePositionAtUnit`
2. Keep the same api(same function name && same parameters)
Disadvantage:
1. Duplicate code

glabute: I think you need both.  CinemachinePathBase had both, if I remember correctly.  We should do what it takes to minimize impact on client classes

jieyingLin: Yea that's what I mean, my changes is add `StandardizeUnit` into `EvaluatePositionAtUnit`, so in `Dolly cart` we only need to call `EvaluatePositionAtUnit`, but in helper class we will have `StandardizeUnit` and `EvaluatePositionAtUnit`.

glabute: 👍

Runtime/Core/UnitySplineExtension.cs
```diff
@@ -0,0 +1,126 @@
...
+            PathUnits,
+            /// <summary>Use Distance Along Path.  Path will be sampled according to its Resolution
+            /// setting, and a distance lookup table will be cached internally</summary>
+            Distance,
+            /// <summary>Normalized units, where 0 is the start of the path, and 1 is the end.
+            /// Path will be sampled according to its Resolution
+            /// setting, and a distance lookup table will be cached internally</summary>
+            Normalized
+        }
+
```
jieyingLin, line 23: I didn't implement the MinPos since we don't need right now

Runtime/Core/CinemachinePathCache.cs
```diff
@@ -0,0 +1,429 @@
...
+        {
+            m_DistanceToPos = null;
+            m_PosToDistance = null;
+            m_CachedSampleSteps = 0;
+            m_PathLength = 0;
+        }
+
+        /// <summary>When calculating the distance cache, sample the path this many
+        /// times between points</summary>
+        private int m_DistanceCacheSampleStepsPerSegment = 20;
```
jieyingLin, line 37: Remove `m_Resolution`

Runtime/Core/CinemachinePathCache.cs
```diff
@@ -0,0 +1,429 @@
...
+                Mathf.Cos(halfAngle));
+        }
+
+        /// <summary>Get the orientation the curve at a point along the path.</summary>
+        /// <param name="pos">Postion along the path.  Need not be standardized.</param>
+        /// <returns>World-space orientation of the path</returns>
+        private Quaternion EvaluateOrientation(float pos)
+        {
+            return Quaternion.identity;
+        }
```
jieyingLin, line 232: Waiting for spline team support for `EvaluateOrientation` and `EvaluateTangent`

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -51,6 +54,15 @@ public enum UpdateMethod
         [FormerlySerializedAs("m_CurrentDistance")]
         public float m_Position;
 
+        private void OnValidate()
```
glabute, line 57: OnValidate is called after changes in the inspector, and after an object is deserialized.  If some user script changes m_path during runtime, then OnValidate won't be called, so you need to be validating this more frequently.  Since the cost of this test is small, it doesn't hurt to validate on each use.

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -59,6 +71,8 @@ void FixedUpdate()
 
         void Update()
         {
+            m_PathCache.InvalidateDistanceCache();
```
glabute, line 74: I don't think you want to invalidate every frame!  That kills the whole value of caching.

glabute (changes requested):

Runtime/Behaviours/CinemachineDollyCart.cs
jieyingLin: I remove `OnValidate` function and move this part of code into `Update()`. Please LMK if there is a better place to place it

Runtime/Behaviours/CinemachineDollyCart.cs
jieyingLin: Yes you are right. I am waiting for callback() from spline team. It is in PR already. After I get it I will do something like 
`if (spline.change) InvalidateDistanceCache ()` here

Runtime/Components/CinemachineTrackedDolly.cs
```diff
@@ -201,20 +204,33 @@ public override void MutateCameraState(ref CameraState curState, float deltaTime
...
                 return;
+            }
+
+            if (m_PathCache.m_Path != m_Path)
+            {
+                m_PathCache.m_Path = m_Path;
+                m_PathCache.InvalidateDistanceCache();
+            }
+
+            m_PathCache.InvalidateDistanceCache();
```
jieyingLin, line 219: Waiting for `callback()` here

Editor/Editors/CinemachineTrackedDollyEditor.cs
```diff
@@ -54,11 +54,10 @@ private static void DrawTrackeDollyGizmos(CinemachineTrackedDolly target, GizmoT
         {
             if (target.IsValid)
             {
-                CinemachinePathBase path = target.m_Path;
+                CinemachinePathCache path = target.m_PathCache;
                 if (path != null)
                 {
                     var isActive = CinemachineCore.Instance.IsLive(target.VirtualCamera);
-                    CinemachinePathEditor.DrawPathGizmo(path, path.m_Appearance.pathColor, isActive);
```
jieyingLin, line 61: I remove the `Appearance` here since I think it should be a part of `SplineContainer`. I will talk to the spline team to see if I can contributes some codes there. If they feel it doesn't make sense to add code only used in `cinemachine` I will be back to create some extensions.

glabute (changes requested):

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -59,6 +62,14 @@ void FixedUpdate()
 
         void Update()
         {
+            if (m_Path && m_PathCache.m_Path != m_Path)
+            {
+                m_PathCache.m_Path = m_Path;
+                m_PathCache.InvalidateDistanceCache();
+            }
```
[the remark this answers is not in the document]
jieyingLin: I move this part into a new function called `UpdatePathCache()` , this function will update the cached path elements when necessary

gaborkb:
TODO: Spline include guards - use dolly track when splines are not present. Splines Package (https://github.com/Unity-Technologies/com.unity.splines) is not going to be backported to Unity Editor versions before 2020.2.

gaborkb:
Converted to draft, until the include guards are not added.

gaborkb:
TODO2: splines team added some functionality for rotation, check if it is relevant:
https://unity.slack.com/archives/C01AAJ9069Z/p1631811838007400

glabute:
SplinePath Inspector is not working for me:
image

glabute:
nvm about the inspectors.  The problem goes away when I downgrade Unity from 0b12 to 0b3

glabute (changes requested):
We need to make sure that our use of splines is conforming to best practices (currently it probably isn't).  For example, the way they did the road.  Our dolly path should be following a similar pattern, I think.

The rotation should probably be implemented using the built-in knot rotation directly.

They have that nice "animate along path" behaviour, which is the same thing as our dolly cart.  

They have much of the functionality of CinemachinePathBase built-in (specifically unit conversion and path length cache).  We should use it and not be duplicating stuff.

We need to figure out how to use their stuff to the fullest extent possible.  It probably means moving beyond Cheryl's naive implementation

One possibility to consider: implement *everything* (including legacy path classes) on top of spline.

Runtime/Behaviours/CinemachineDollyCart.cs
```diff
@@ -75,6 +75,9 @@ void SetCartPosition(float distanceAlongPath)
                 m_Position = m_Path.StandardizeUnit(distanceAlongPath, m_PositionUnits);
                 transform.position = m_Path.EvaluatePositionAtUnit(m_Position, m_PositionUnits);
                 transform.rotation = m_Path.EvaluateOrientationAtUnit(m_Position, m_PositionUnits);
+            } else
+            {
+                m_Position = 0;
```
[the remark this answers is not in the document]
glabute: OR... bite the bullet and leave it as a hard dependency.  Might make sense if we re-implement legacy classes in terms of spline

PR #106, opened 2020-09-28
Bugfix for 1174993 + release notes

gaborkb:
Fix for https://fogbugz.unity3d.com/f/cases/1174993/.

The bug was caused by 
```
AssetDatabase.LoadAssetAtPath<Texture2D>( ScriptableObjectUtility.CinemachineRealativeInstallPath + "/Editor/EditorResources/cm_logo_sm.png"); 
```
Because AssetDatabase does not find the Texture after the package is added. We need to wait one update cycle.


glabute (approved):
I fixed the spacing - somehow you inserted tabs instead of spaces.  Do you need to fix your editor settings?

PR #107, opened 2020-09-29
Dev/standardized ux

gaborkb:
Moved Cinemachine menu to GameObject.
Removed About page.


JulienDidierjean:
Hi Krisztian, 
Thanks for putting me as a reviewer...
Here are my comments based on the screenshot you posted:
 - I would drop the "Create" from each line, none of the the other entries in this "Create" menu has one
 - I'm quite sure I would not put an enable/disable option there: "Storyboard Global Mute" 
I think we should put it under Cinemachine Preferences maybe... 
Question: Is this option not present in the UI of all the Cinemachine cameras ? Maybe it's enough...
Can you tell me why it's been added there in the first place ?
(I can post this comment in the cinemachine-core channel to trigger discussions if you want...

gaborkb:
> * I would drop the "Create" from each line, none of the the other entries in this "Create" menu has one

Ok!

gaborkb:
> * I'm quite sure I would not put an enable/disable option there: "Storyboard Global Mute"
>   I think we should put it under Cinemachine Preferences maybe...
>   Question: Is this option not present in the UI of all the Cinemachine cameras ? Maybe it's enough...
>   Can you tell me why it's been added there in the first place ?
>   (I can post this comment in the cinemachine-core channel to trigger discussions if you want...

Yes, please post it, as I am not sure what the answer is.

Editor/Editors/CinemachineStoryboardEditor.cs
```diff
@@ -11,7 +11,7 @@ namespace Cinemachine.Editor
     [InitializeOnLoad]
     internal class CinemachineStoryboardMute
     {
-        const string StoryboardGlobalMuteMenuName = "Cinemachine/Storyboard Global Mute";
+        const string StoryboardGlobalMuteMenuName = "GameObject/Cinemachine/Storyboard Global Mute";
```
glabute, line 14: Leave this in the Cinemachine menu for now (maybe as the only thing).
We need to create a preferences page to move it to.  We can save that for another PR.
(so just revert this change)

Extras~/ReleaseNotes.txt
```diff
@@ -6,6 +6,7 @@
 - Bugfix (1256530) - disallow multiple components where appropriate
 - Bugfix: BlendList camera was incorrectly holding 0-length camera cuts
 - Bugfix (1174993) - CM Brain logo was not added to Hierarchy next to Main Camera after adding vcam for the first time after importing CM. 
+- UI update - Moved Cinemachine menu to GameObject Create menu and Right Click context menu for Hierachy window.
```
glabute, line 9: This entire file can now be deleted, since it was only ever used in the About window.

glabute (changes requested):
You also need to go through the manual (the Documentation~ folder) and correct any references to the Cinemachine menu

glabute (approved):

PR #117, opened 2020-10-16
Proposal for an alternative corner damping algorithm

glabute:
This way of doing corner damping seems cleaner to me.  What do you think?


glabute:
oops - this PR is wrong.  Closing

PR #118, opened 2020-10-16
Proposal for alternative corner damping

glabute:
I think it's simpler and cleaner this way, and no popping edge cases, ever.
What do you think?


gaborkb (commented):
Merged this locally into my branch, so we can close this one.

PR #132, opened 2020-11-12
Unit test for checking unity camera position vs vcam position

jieyingLin:


gaborkb (approved):
Very nice and clean structure

PR #133, opened 2020-11-13
Fogbugz

bryanzakounity:
Changes to doc re: fogbugz tickets. Small minor mods.


glabute (changes requested):
DS_Store should not be included in the repo.  Please remove it.

PR #134, opened 2020-11-17
Dev/confiner2 d find intersection

glabute:
Fixed FindIntersection to handle overlapping and zero-length lines.
After implementing this, I noticed that a confiner unity test is failing.  It's possible that the old behaviour of FindIntersection was hiding a bug.


glabute:
ack! did it again!!!!!!

PR #135, opened 2020-11-17
Fix FindIntersection to handle parallel segments

glabute:
Fixed FindIntersection to handle overlapping and zero-length segments.
After implementing this, I noticed that a confiner unity test is failing. It's possible that the old behaviour of FindIntersection was hiding a bug.

### DIFF
--- a/Runtime/Behaviours/CinemachineConfiner2D.cs
+++ b/Runtime/Behaviours/CinemachineConfiner2D.cs
@@ -270,9 +270,8 @@ namespace Cinemachine
                 {
                     Vector2 p1 = polygon[index];
                     Vector2 p2 = polygon[(index + 1) % polygon.Count];
-                    int intersectionType = UnityVectorExtensions.FindIntersection(p, camRayEndFromCamPos2D, p1, p2, 
-                        out var intersection);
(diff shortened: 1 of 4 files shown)


Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -271,8 +271,8 @@ static bool IsInside(in List<List<Vector2>> polygons, in Vector2 p)
                     Vector2 p1 = polygon[index];
                     Vector2 p2 = polygon[(index + 1) % polygon.Count];
                     int intersectionType = UnityVectorExtensions.FindIntersection(p, camRayEndFromCamPos2D, p1, p2, 
-                        out _);
-                    if (intersectionType == 2)
+                        out var intersection);
+                    if (intersectionType == 2 && (intersection - p2).sqrMagnitude < 0.01f)
```
gaborkb, line 275: There is no need to check for (intersection - p2).sqrMagnitude < 0.01f here. 
I checked and for me the camera is never Inside with this extra condition. If I remove it, then it correctly identifies inside/outside.

gaborkb (changes requested):

gaborkb (approved):

PR #139, opened 2020-11-25
Dev/confiner2 d clipper refactor

gaborkb:
I was thinking that maybe we could add a Time limit to BakeConfiner. Because it is possible to give a huge-huge polygon (lot of points) which will choke up the editor.
For example, 10 seconds. If BakeConfiner needs to stop because of time limit then:
- Display warning message to user: "Confiner timed out. Reduce the number of points in your polygon, or to increase computation time, set "parameter"...
- Don't recompute, until invalidateCache is triggered.


Runtime/Core/ConfinerOven.cs
```diff
@@ -42,156 +263,146 @@ public float MaxFrustumHeight
...
-                    leftCandidate = splitPoly;
-                    rightCandidate = null;
+                    leftCandidate = rightCandidate;
+                    rightCandidate = new PolygonSolution();
                     
                     // Back to max step.  GML todo: this can be smaller now
                     stepSize = maxStepSize;
                 }
-                else if (rightCandidate == null && leftCandidate[0].m_FrustumHeight >= maxFrustumHeight)
+                else if (rightCandidate.IsEmpty && leftCandidate.m_FrustumHeight >= maxFrustumHeight)
```
glabute, line 381: should this be || instead of && ?

glabute (changes requested):

PR #140, opened 2020-11-26
Vcam priority set from script did not trigger resort of active vcams.

gaborkb:
https://fogbugz.unity3d.com/f/cases/1293429/
The problem is that we start with three vcams: A, B, C.
A priority = 30
B priority = 20
C priority = 10

From script we change the priorities to A=0, B=0, C=10. Still, brain selects camera A, because that is at the top of the priority list (because it was not resorted).


gaborkb:
Do we need MoveToTopOfPrioritySubqueue and Mode.PriorityBoost after the sort change? I think, it is not needed anymore, or is it?

PR #145, opened 2021-01-13
Storyboard 1 pixel border bug

gaborkb:
The bug was caused because the viewport size was one pixel shorter.

The bug is only present with not-pixel-perfect setups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,6 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [2.7.3]
-- Bugfix: CM StoryBoard had a 1 pixel border.
-
 ## [2.7.2] - 2020-12-31
 - CinemachineConfiner2D now handles cases where camera window is oversized
 - Bugfix (1293429) - Brain could choose vcam with not the highest priority in some cases.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.7.3]
```
glabute, line 7: we haven't released 2.7.2 yet, move to that section

Runtime/Behaviours/CinemachineStoryboard.cs
```diff
@@ -251,7 +251,7 @@ void PlaceImage(CanvasInfo ci, float alpha)
                 ci.mViewport.localRotation = Quaternion.identity;
                 ci.mViewport.localScale = Vector3.one;
                 ci.mViewport.ForceUpdateRectTransforms();
-                ci.mViewport.sizeDelta = new Vector2(screen.width - Mathf.Abs(wipeAmount), screen.height);
+                ci.mViewport.sizeDelta = new Vector2(screen.width + 1 - Mathf.Abs(wipeAmount), screen.height + 1);
```
glabute, line 254: Can you prove to me that precisely-sized images won't get stretched?  i.e. we can do pixel-perfect storyboards

glabute (changes requested):

PR #149, opened 2021-01-19
New sample scene: FadeOut

gaborkb:
Added a sample scene with a fadeout material example. Objects fade out in front of the camera.

While tidying up the project, I thought maybe we could add a view angle to the shader.


glabute:
I think you branched this off your analytics branch by mistake.  You might want to make a new one

PR #150, opened 2021-01-19
FadeOut SampleScene

gaborkb:
Two fadeout sample scenes.

FadeOut sample scene with a cinemachine extension that simplifies control over the shader. The sample scene has its own material and shader in the same folder, because these are not shared. The cinemachine extension is in the example package, because without the shader it is useless.


glabute:
Fix only this, and it's perfect.

image

PR #151, opened 2021-01-21
CMCL-207 bugfix: create from menu must create as child object

glabute:
bugfix for CMCL-207


Editor/Menus/CinemachineMenu.cs
```diff
@@ -36,40 +36,49 @@ private static void CreateFixedSignalDefinition()
         private const string m_CinemachineGameObjectRootMenu = "GameObject/Cinemachine/";
         private const int m_MenuPriority = 11; // right after Camera
         
+        static void SetParentToMenuContextObject(GameObject child, MenuCommand command)
+        {
+            var go = command.context as GameObject;
+            var parent = go == null ? null : go.transform;
+            if (parent != null)
+                Undo.SetTransformParent(child.transform, parent, "set parent");
```
gaborkb, line 44: Why not just:
```
var go = command.context as GameObject;
if (go != null && go.transform != null)
     Undo.SetTransformParent(child.transform, go.transform, "set parent");
```

Editor/Menus/CinemachineMenu.cs
```diff
@@ -36,40 +36,49 @@ private static void CreateFixedSignalDefinition()
...
-            return InternalCreateVirtualCamera(
-                "CM vcam", true, typeof(CinemachineComposer), typeof(CinemachineTransposer));
+            var go = InternalCreateVirtualCamera(
+                "CM vcam", false, typeof(CinemachineComposer), typeof(CinemachineTransposer)).gameObject;
+            SetParentToMenuContextObject(go, command);
         }
 
         [MenuItem(m_CinemachineGameObjectRootMenu + "FreeLook Camera", false, m_MenuPriority)]
-        private static void CreateFreeLookCamera()
+        private static void CreateFreeLookCamera(MenuCommand command)
```
gaborkb, line 57: New dots guide says that we should not write redundant modifiers, so no privates.

If we don't follow that in classic cm, then at line 48, we also need to have private for consistency.

gaborkb (changes requested):

gaborkb (approved):

PR #152, opened 2021-01-21
[CMCL-208] Bugfix: FreeLook small drift when no user input if SimpleFollow

glabute:
fix for CMCL-208


gaborkb (approved):
Don't merge it to master until Patrick looks at it. :)

PR #157, opened 2021-02-04
Prototype: Blend start and end callbacks on Brain

gaborkb:
A user requested access to blend start and end events. (https://jira.unity3d.com/browse/CMCL-34, https://forum.unity.com/threads/faster-than-camera-main.841246/#post-5555557) 
I thought adding a callback would be a good idea.

This is a very simple prototype. I did not want to implement the feature fully without agreeing on it with you. (so no editor scripts, not fully thought out, did not test with timeline, etc).

I found the way I check for blend start and blend end works in a simple scene with a couple of vcams. But maybe there is a better way of checking and wiring the callbacks to the blend. Maybe the start and end callbacks should be in brain and not passed to blend definition. Not sure.

What do you think?

### DIFF
--- a/Runtime/Behaviours/CinemachineBrain.cs
+++ b/Runtime/Behaviours/CinemachineBrain.cs
@@ -164,6 +164,18 @@ namespace Cinemachine
         /// The parameters are (incoming_vcam, outgoing_vcam), in that order.
         /// </summary>
         [Serializable] public class VcamActivatedEvent : UnityEvent<ICinemachineCamera, ICinemachineCamera> {}
+        
+        /// <summary>
+        /// Event that is fired when a blend has started.
+        /// The parameters are (incoming_vcam, outgoing_vcam), in that order.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


krisrok:
Upon further inspection: Is there a specific reason for passing the event references into the blend instances? The brain is the only one updating the blends (at least the ones used here) so it already knows when a blend completed if I'm not mistaken.

Invoke the Start event somewhere around here: https://github.com/Unity-Technologies/com.unity.cinemachine/blob/36746dec85fcd496805d432bc0e245f8573ee5b2/Runtime/Behaviours/CinemachineBrain.cs#L754-L783

And the End event here: https://github.com/Unity-Technologies/com.unity.cinemachine/blob/36746dec85fcd496805d432bc0e245f8573ee5b2/Runtime/Behaviours/CinemachineBrain.cs#L793-L800

What do you think?

(Cuts would need special handling and I'm not a fan of how they are handled right now.)

glabute:
Implemented in https://github.com/Unity-Technologies/com.unity.cinemachine/pull/844

PR #158, opened 2021-02-05
Input system enabled/disabled

gaborkb:
https://forum.unity.com/threads/input-action-reference-wont-disable-on-cinemachine.1052192/

We don't check whether Player Input is enabled, before using it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: CM StoryBoard lost viewport reference after hot reload
 - Bugfix: FramingTransposer's TargetMovementOnly damping caused a flick.
 - Bugfix: InheritPosition did not work with SimpleFollow binding mode
+- Bugfix: Checking whether the Input Action is enabled before using it.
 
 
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: CM StoryBoard lost viewport reference after hot reload
 - Bugfix: FramingTransposer's TargetMovementOnly damping caused a flick.
 - Bugfix: InheritPosition did not work with SimpleFollow binding mode
+- Bugfix: Checking whether the Input Action is enabled before using it.
```
glabute, line 20: you should mention CinemachineInputHandler here, to contextualize

Runtime/Helpers/CinemachineInputProvider.cs
```diff
@@ -69,6 +69,9 @@ public virtual float GetAxisValue(int axis)
         /// <returns>The cached action for the player specified in PlayerIndex</returns>
         protected InputAction ResolveForPlayer(int axis, InputActionReference actionRef)
         {
+            if (!actionRef.action.enabled)
```
glabute, line 72: what if actionRef or actionRef.action is null?

glabute (changes requested):

glabute (approved):

PR #163, opened 2021-02-09
Don't use scriptable object mono path to get CM install path

mjmvisser:
Just assume we're in a package and use straightforward relative path resolution. Drops the non-upm package support based on substring manipulation.
Also added unit tests.
Fixes CMCL-223.


glabute:
You only pushed the tests, what about the bugfix?

glabute (approved):

PR #164, opened 2021-02-09
CMCL-197: add damping to 3rdPersonFollow collision handling

gaborkb:
Adds pull back damping after a collision.

I am damping the distance to make the pull back rotationally invariant, that is no damping on user rotation. Just like the rest of the third person rig.

https://jira.unity3d.com/browse/CMCL-197

### DIFF
--- a/Runtime/Components/Cinemachine3rdPersonFollow.cs
+++ b/Runtime/Components/Cinemachine3rdPersonFollow.cs
@@ -122,7 +122,8 @@ namespace Cinemachine
 
         /// <summary>Orients the camera to match the Follow target's orientation</summary>
         /// <param name="curState">The current camera state</param>
-        /// <param name="deltaTime">Not used.</param>
+        /// <param name="deltaTime">Elapsed time since last frame, for damping calculations.  
+        /// If negative, previous state is reset.</param>
         public override void MutateCameraState(ref CameraState curState, float deltaTime)
         {
             if (IsValid)
--- a/Runtime/Components/Cinemachine3rdPersonFollow.cs
+++ b/Runtime/Components/Cinemachine3rdPersonFollow.cs
@@ -80,12 +80,12 @@ namespace Cinemachine
         public float PostCorrectionDamping = 0;
 
         // State info
-        private Vector3 PreviousFollowTargetPosition;
-        private float PreviousHeadingAngle;
-        float m_prevHandDistance;
-        float m_prevCamPosDistance;
+        Vector3 m_PreviousFollowTargetPosition;
+        float m_PreviousHeadingAngle;
+        float m_PrevHandDistance;
+        float m_PrevCamPosDistance;
 
-        private void OnValidate()
+        void OnValidate()
         {
             CameraSide = Mathf.Clamp(CameraSide, -1.0f, 1.0f);
             Damping.x = Mathf.Max(0, Damping.x);
@@ -94,7 +94,7 @@ namespace Cinemachine
             CameraRadius = Mathf.Max(0.001f, CameraRadius);
         }
 
-        private void Reset()
+        void Reset()
         {
             CameraCollisionFilter = 1;
             ShoulderOffset = new Vector3(0.5f, -0.4f, 0.0f);
@@ -106,7 +106,7 @@ namespace Cinemachine
         }
 
         /// <summary>True if component is enabled and has a Follow target defined</summary>
-        public override bool IsValid { get { return enabled && FollowTarget != null; } }
+        public override bool IsValid => enabled && FollowTarget != null;
 
         /// <summary>Get the Cinemachine Pipeline stage that this component implements.
         /// Always returns the Aim stage</summary>
@@ -132,7 +132,7 @@ namespace Cinemachine
 
         void PositionCamera(ref CameraState curState, float deltaTime)
         {
-            var prevTargetPos = deltaTime >= 0 ? PreviousFollowTargetPosition : FollowTargetPosition;
+            var prevTargetPos = deltaTime >= 0 ? m_PreviousFollowTargetPosition : FollowTargetPosition;
 
             // Compute damped target pos (compute in camera space)
             var dampedTargetPos = Quaternion.Inverse(curState.RawOrientation) 
@@ -149,16 +149,16 @@ namespace Cinemachine
             var followTargetForward = followTargetRotation * fwd;
             var angle = UnityVectorExtensions.SignedAngle(
                 fwd, followTargetForward.ProjectOntoPlane(up), up);
-            var previousHeadingAngle = deltaTime >= 0 ? PreviousHeadingAngle : angle;
+            var previousHeadingAngle = deltaTime >= 0 ? m_PreviousHeadingAngle : angle;
             var deltaHeading = angle - previousHeadingAngle;
-            PreviousHeadingAngle = angle;
+            m_PreviousHeadingAngle = angle;
 
             // Bypass user-sourced rotation
             dampedTargetPos = FollowTargetPosition 
                 + Quaternion.AngleAxis(deltaHeading, up) * (dampedTargetPos - FollowTargetPosition);
-            PreviousFollowTargetPosition = dampedTargetPos;
+            m_PreviousFollowTargetPosition = dampedTargetPos;
 
-            GetRigPositions(out Vector3 root, out Vector3 shoulder, out Vector3 hand);
+            GetRigPositions(out Vector3 root, out _, out Vector3 hand);
 
             // 1. Check if pivot itself is colliding with something, if yes, then move the pivot 
             // closer to the player. The radius is bigger here than in step 2, to avoid problems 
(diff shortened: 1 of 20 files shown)


Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -121,12 +134,11 @@ public override void MutateCameraState(ref CameraState curState, float deltaTime
 
         void PositionCamera(ref CameraState curState, float deltaTime)
         {
-            var targetPos = FollowTargetPosition;
```
glabute, line 124: FollowTargetPosition is a function call.  It makes sense to grab it here instead of calling it multiple times.  This cleanup is too zealous!

This drives home what Cort was saying in the last DOTS townhall.  Function calls that look like simple variable reads are misleading and dangerous.

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -92,10 +104,11 @@ private void Reset()
             CameraDistance = 2.0f;
             Damping = new Vector3(0.1f, 0.5f, 0.3f);
             CameraRadius = 0.2f;
+            PostCorrectionDamping = 2f;
```
glabute, line 107: The name feels awkward.  How about "CollisionDamping"?

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -139,56 +151,96 @@ void PositionCamera(ref CameraState curState, float deltaTime)
...
-            camPos = PullTowardsStartOnCollision(in hand, in camPos, in CameraCollisionFilter, CameraRadius);
-
-            curState.RawPosition = camPos;
+            Vector3 camPos = dampedHandResolved - (followTargetForward * CameraDistance);
+            bool camPosHasHit = PullTowardsStartOnCollision(in dampedHandResolved, in camPos, in CameraCollisionFilter, 
+                CameraRadius, out var camPosResolved);
+            // Post correction damping
+            Vector3 dampedCamPosResolved = DampedPullBackPostCollision(camPosHasHit, deltaTime, 
+                    root, camPosResolved, ref m_PrevCamPosDistance);
+
```
glabute, line 182: This will also introduce damping if user changes the camera distance through script, and we don't want that.  We should only damp when returning from collision.

You can test that in the AimingRig sample scene by enabling the optional ThirdPersonFollowDistanceModifier on the vcam.  It scales the distance as a function of vertical angle, to mimic the 3 rigs of the FreeLook.

glabute (changes requested):

glabute:
Also: don't forget to add a line in the user manual.  Bryan is still writing that so it isn't available here, but we must remember to do it.

Runtime/Components/Cinemachine3rdPersonFollow.cs
gaborkb: OMG fell into that trap right away. Thanks!!!

Runtime/Components/Cinemachine3rdPersonFollow.cs
gaborkb: Pushed fix. Now we only damp post collision, until camera is back at its intended position.

What should happen if the user is modifying the intended position while we damp here? E.g. they look down and ThirdPersonFollowDistanceModifier is active, then currently we damp that extra distance too.

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -69,21 +69,39 @@ public class Cinemachine3rdPersonFollow : CinemachineComponentBase
...
+        Vector3 m_PreviousFollowTargetPosition;
+        float m_PreviousHeadingAngle;
+        
+        float m_PrevHandDistance;
+        bool m_PrevHandHasHit;
+        bool m_IsHandPostCollision;
+        
+        bool m_PrevCamPosHasHit;
+        bool m_IsCamPosPostCollision;
+        float m_PrevCamPosDistance;
```
glabute, line 92: This is way more complex than it needs to be.  Don't need all this state

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -139,56 +158,103 @@ void PositionCamera(ref CameraState curState, float deltaTime)
...
+            bool camPosHasHit = PullTowardsStartOnCollision(in dampedHandResolved, in camPos, in CameraCollisionFilter, 
+                CameraRadius, out var camPosResolved);
+            // Post collision damping
+            Vector3 dampedCamPosResolved = DampedPullBackPostCollision(camPosHasHit, deltaTime, root, camPosResolved, 
+                ref m_PrevCamPosHasHit, ref m_IsCamPosPostCollision, ref m_PrevCamPosDistance);
 
-            curState.RawPosition = camPos;
+            // Set state
+            curState.RawPosition = dampedCamPosResolved;
             curState.RawOrientation = FollowTargetRotation;
```
glabute, line 192: should be followTagetPosition (not your fault)

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -121,12 +140,12 @@ public override void MutateCameraState(ref CameraState curState, float deltaTime
 
         void PositionCamera(ref CameraState curState, float deltaTime)
         {
-            var targetPos = FollowTargetPosition;
-            var prevTargetPos = deltaTime >= 0 ? PreviousFollowTargetPosition : targetPos;
+            var followTarget = FollowTargetPosition;
+            var prevTargetPos = deltaTime >= 0 ? m_PreviousFollowTargetPosition : followTarget;
```
glabute, line 144: all the "prev" states need to be reset when deltaTime < 0, including the new ones just added

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -139,56 +158,103 @@ void PositionCamera(ref CameraState curState, float deltaTime)
...
+            if (postCollision && CollisionDamping > 0 && deltaTime >= 0)
+            {
+                Vector3 difference = resolved - root;
+                float delta = difference.magnitude - previousDistance;
+                delta = Damper.Damp(delta, CollisionDamping, deltaTime);
+                dampedResolved = root + difference.normalized * (delta + previousDistance);
+
+                postCollision = (resolved - dampedResolved).sqrMagnitude >= UnityVectorExtensions.Epsilon;
+            }
+            previousDistance = (dampedResolved - root).magnitude;
```
glabute, line 256: this code is very inefficient - lots of sqrts being calculated, when you only need to do it once

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -139,56 +158,103 @@ void PositionCamera(ref CameraState curState, float deltaTime)
...
 
             // 2. Try to place the camera to the preferred distance
-            var camPos = hand - (followTargetForward * CameraDistance);
-            camPos = PullTowardsStartOnCollision(in hand, in camPos, in CameraCollisionFilter, CameraRadius);
+            Vector3 camPos = dampedHandResolved - (followTargetForward * CameraDistance);
+            bool camPosHasHit = PullTowardsStartOnCollision(in dampedHandResolved, in camPos, in CameraCollisionFilter, 
+                CameraRadius, out var camPosResolved);
+            // Post collision damping
+            Vector3 dampedCamPosResolved = DampedPullBackPostCollision(camPosHasHit, deltaTime, root, camPosResolved, 
+                ref m_PrevCamPosHasHit, ref m_IsCamPosPostCollision, ref m_PrevCamPosDistance);
```
glabute, line 188: you have introduced a lot of extra variables here, not sure why it's needed.  It's confusing.

glabute (changes requested):
Krisztian, I understand that this is a big day for you, so I'm going to make the fixes myself.  Please review it when you have time :)

Runtime/Components/Cinemachine3rdPersonFollow.cs
glabute: I think we have no choice but to damp that extra distance

glabute:
Adding @mjmvisser as reviewer for my coomit

gaborkb (commented):
Great optimization and clean up Gregory! 👍

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -112,21 +125,27 @@ private void Reset()
         /// <param name="deltaTime">Not used.</param>
```
mjmvisser, line 125: "Not used" but 
```
if (!VirtualCamera.PreviousStateIsValid)
    deltaTime = -1;
```

mjmvisser (approved):
looks good to me - see comment about deltaTime param in `MutateCameraState`

metalpat (approved):
Tested collision damping in Unity 2020.2.3f1. Verified change in Sample Scene as well

mjmvisser (approved):

mjmvisser (approved):

PR #171, opened 2021-02-10
Toggle editable/non-editable game window guides

gaborkb:
### Purpose of this PR

Added Toggle UI in Cinemachine Preferences. Users requested an option to turn off Editable Game Window Guides. https://jira.unity3d.com/browse/CMCL-222 (-> https://forum.unity.com/threads/non-editable-game-window-guide.1054085/)

### Testing status

- [N/A] Added an automated test
- [N/A] Passed all automated tests
- [X] Manually tested 

### Documentation status

- [X] Updated [CHANGELOG](https://keepachangelog.com/en/1.0.0/)
- [N/A] Updated README (if applicable)
- [X] Commented all public classes, properties, and methods

### Technical risk

0

### Package version
- [ ] Updated package version


glabute (changes requested):
This needs to be an editor preference so that the setting gets saved.  See for example the implementation of CinemachineStoryboardMute

Editor/Editors/CinemachineScreenComposerGuidesGlobalEditable.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using UnityEditor;
+
+namespace Cinemachine.Editor
+{
+    [InitializeOnLoad]
+    class CinemachineScreenComposerGuidesGlobalEditable // why not static? (same Q for CinemachineStoryboardMute)
```
gaborkb, line 7: Should I move this in Editor/Editors/CinemachineScreenComposerGuides.cs?

glabute: yes please.
And yes, I suppose it could be a static class.  Can you find out if it makes a difference, and explain to me what that difference is?

Editor/Windows/CinemachineSettings.cs
```diff
@@ -329,7 +329,13 @@ private static void OnGUI()
                 new GUIContent("Storyboard Global Mute", "If checked, all storyboards are globally muted."), 
                 CinemachineStoryboardMute.Enabled);
 #endif
-
+            // Editable Game Window Guides
+            CinemachineScreenComposerGuidesGlobalEditable.Enabled = EditorGUILayout.Toggle(
+                new GUIContent("Editable Game Window Guides", "If checked, game window guides are editable in play " +
```
glabute, line 334: For me, it's cut off.  Please fix.

image

glabute (changes requested):

Editor/Editors/CinemachineScreenComposerGuidesGlobalEditable.cs
gaborkb: **Reason for Static class**
- "A static class is basically the same as a non-static class, but there is one difference: a static class cannot be instantiated." [1]
- Static class is more restrictive than a normal class. Both classes (CMStoryboardMute and CMScreenComposerGuidesEditable) fit into this more restrictive rule set. 

The more restrictions are:
1. contains only static members,
2. cannot be instantiated,
3. is sealed, and
4. cannot contain Instance Constructors. [1]

**Reason for InitializeOnLoad attribute and static constructor:**
-  "A static constructor is used to initialize any static data, or to perform a particular action that needs to be performed once only. It is called automatically before the first instance is created or any static members are referenced." [1]
- Static constructors with InitializeOnLoad attribute are called when scripts in the project are recompiled. [2]

[1] https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-classes-and-static-class-members
[2] https://docs.unity3d.com/ScriptReference/InitializeOnLoadAttribute.html

Editor/Windows/CinemachineSettings.cs
```diff
@@ -323,12 +323,23 @@ private static void OnGUI()
                 GUILayout.EndScrollView();
             }
 
+            // set label width, so text is not cut for Toggles
+            float originalValue = EditorGUIUtility.labelWidth;
+            EditorGUIUtility.labelWidth = 180f;
```
glabute, line 328: The solution is good, but I don't like this hardcoded number.  It's a problem waiting to happen.
Instead, you should get the number dynamically using GUI.skin.label.CalcSize.  See examples elsewhere in code.  Calc size for both labels and use the max.

Editor/Windows/CinemachineSettings.cs
```diff
@@ -323,12 +323,23 @@ private static void OnGUI()
...
-                new GUIContent("Storyboard Global Mute", "If checked, all storyboards are globally muted."), 
-                CinemachineStoryboardMute.Enabled);
+                // Storyboard global mute
+                CinemachineStoryboardMute.Enabled = EditorGUILayout.Toggle(
+                    new GUIContent("Storyboard Global Mute", "If checked, all storyboards are globally muted."),
+                    CinemachineStoryboardMute.Enabled);
 #endif
+                // Editable Game Window Guides 
+                CinemachineScreenComposerGuidesGlobalEditable.Enabled = EditorGUILayout.Toggle(
+                    new GUIContent("Editable Game Window Guides", "If checked, game window guides are editable in play " +
```
glabute, line 338: I think it should be "Draggable" instead of "Editable", because you can still edit it from the inspector.

glabute (changes requested):

glabute:
image

Also fix tooltip, it should say "draggable" not "editable".

glabute (approved):

PR #215, opened 2021-03-30
SaveDuringPlay only saves SaveDuringPlay components

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-283
SaveDuringPlay saves everything that is on a vcam's gameobject, which is not desired. 

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

Low

### Package version

- [ ] Updated package version


gaborkb:
I saw that CinemachineExtensions don't have [SaveDuringPlay]. I added it for the parent class (CinemachineExtension), so all extensions now [SaveDuringPlay]. However, I saw CinemachineComponentBase does not have [SaveDuringPlay], instead some (possibly all) the subclasses have [SaveDuringPlay].
Should I do the same for extensions?

Runtime/Core/CinemachineExtension.cs
```diff
@@ -9,6 +9,7 @@ namespace Cinemachine
     /// to the vcam, modifying its generated state
     /// </summary>
     [DocumentationSorting(DocumentationSortingAttribute.Level.API)]
+    [SaveDuringPlay]
```
glabute, line 12: Don't do this.  It's up to each extension to decide if it wants SaveDuringPlay

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: changing targets on a live vcam was misbehaving
 - Bugfix: Framing transposer now handles empty groups
 - Bugfix: Interrupting a transition with InheritPosition enabled was broken
+- Bugfix: SaveDuringPlay only components that have SaveDuringPlay. Previously it saved all components on a gameobject with at least one component with SaveDuringPlay.
```
glabute, line 25: This is the correct approach - but is it the best implementation?

Editor/Utility/SaveDuringPlay.cs
```diff
@@ -221,24 +221,21 @@ public bool ScanFields(string fullName, MonoBehaviour b)
         /// Recursively scan the MonoBehaviours of a GameObject and its children.
         /// For each leaf field found, call the OnFieldValue delegate.
         /// </summary>
-        public bool ScanFields(GameObject go, string prefix = null)
```
glabute, line 224: I think you should revert this change and go back to your earlier implementation, with this difference: instead of checking whether the MonoBehaviour is a Cinemachine class, check if it has the SaveDuringPlay attribute.

glabute (changes requested):

glabute (approved):

glabute:
if (attr.GetType().Name.Contains("SaveDuringPlay")) 

seems a little weak as a test.  Probably better to test for the exact type of the attribute.
(I realize you just took this snippet from elsewhere in the file - it should probably be fixed in both places)

PR #218, opened 2021-03-31
SaveDuringPlay - Test type string equality instead of contains

gaborkb:

### Purpose of this PR

Made SaveDuringPlay string with equality instead of contains.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version

### DIFF
--- a/Editor/Utility/SaveDuringPlay.cs
+++ b/Editor/Utility/SaveDuringPlay.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using Cinemachine;
 using UnityEditor;
 using UnityEngine;
 
@@ -244,7 +245,7 @@ namespace SaveDuringPlay
             var attrs = b.GetType().GetCustomAttributes(true);
             foreach (var attr in attrs)
             {
-                if (attr.GetType().Name.Contains("SaveDuringPlay"))
+                if (attr is SaveDuringPlayAttribute)
                 {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


glabute (approved):

gaborkb:
Approving your changes. I manually tested yours, and then also manually tested my changes.

Editor/Utility/SaveDuringPlay.cs
```diff
@@ -557,7 +551,6 @@ static void RestoreAllInterestingStates()
                     Undo.RegisterFullObjectHierarchyUndo(go, "SaveDuringPlay");
                     if (saver.PutFieldValues(go, roots))
                     {
-                        //Debug.Log("SaveDuringPlay: updated settings of " + saver.ObjetFullPath);
```
glabute, line 558: actually, I'd leave this line in (but commented out).  Very useful for debugging, and I've had occasion to uncomment it several times

Editor/Utility/SaveDuringPlay.cs
```diff
@@ -385,28 +383,19 @@ static object LeafObjectFromString(Type type, string value, GameObject[] roots)
...
+                    return c == null ? string.Empty : ObjectTreeUtil.GetFullName(c.gameObject);
+                case GameObject go:
+                    return ObjectTreeUtil.GetFullName(go);
+                case ScriptableObject so:
+                    return AssetDatabase.GetAssetPath(so);
+                default:
+                    return obj.ToString();
             }
-            return obj.ToString();
         }
```
glabute: I'm uneasy about this change.  I'm wondering if it properly handles Unity's overriding of the == null operator (which will return true for non-null objects that have been deleted).  I'm doubtful that the c# hocus-pocus here will handle those cases the same way.

glabute (changes requested):

Editor/Utility/SaveDuringPlay.cs
gaborkb: Ok. I checked and it looked like it handles "null" cases too, but to be sure I reverted it back to the original state.

glabute (approved):

PR #219, opened 2021-03-31
Damping by pass applied to previous displacement

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-118
Described in this forum post: https://forum.unity.com/threads/problem-with-cinemachine-collider.948316/

The problem is that collider's displacement is in world space, and when damping is on and the FreeLook rotates, it does not also rotate the offset that's damping itself away.

Solutions could be either: keep offset in camera-local space, or **apply state's damping bypass to the offset**.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version


Runtime/Behaviours/CinemachineCollider.cs
```diff
@@ -325,7 +325,8 @@ public override float GetMaxDampTime()
                     {
                         Vector3 delta = displacement - extra.m_previousDisplacement;
                         delta = Damper.Damp(delta, damping, deltaTime);
-                        displacement = extra.m_previousDisplacement + delta;
+                        displacement = 
+                            Quaternion.Euler(state.PositionDampingBypass) * extra.m_previousDisplacement + delta;
```
glabute, line 329: This looks fishy.  You use an unrotated prev displacement in the first line.  Shouldn't you rotate it first?

glabute (changes requested):
It's better, but I don't think this fix does the whole job.  I can still get it to move laterally in this situation.  See the big horizontal displacement of the vcam (yellow line).  If I then rotate to an unobstructed place, there is slow horizontal damping back to normal

image

gaborkb:
> It's better, but I don't think this fix does the whole job. I can still get it to move laterally in this situation. See the big horizontal displacement of the vcam (yellow line). If I then rotate to an unobstructed place, there is slow horizontal damping back to normal
> 
> image

I think, we should fix that in another PR. Because it think it will be a feature change, not a bug fix.

glabute:
ok, we can do it in a second pass

glabute (approved):

PR #224, opened 2021-04-07
Dev/physical camera override bug

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-288
https://forum.unity.com/threads/aspect-ratios-for-all-my-cameras-got-screwed-up-after-updating-to-unity-2021-1.1088641/

Physical camera parameters should be only overwritten when Override mode: physical is selected.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

0

### Package version

- [ ] Updated package version

### DIFF
--- a/Runtime/Behaviours/CinemachineBrain.cs
+++ b/Runtime/Behaviours/CinemachineBrain.cs
@@ -873,7 +873,7 @@ namespace Cinemachine
                     cam.lensShift = state.Lens.LensShift;
                     cam.orthographic = state.Lens.Orthographic;
                     cam.usePhysicalProperties = state.Lens.IsPhysicalCamera;
-                    if (state.Lens.IsPhysicalCamera)
+                    if (state.Lens.IsPhysicalCamera && state.Lens.ModeOverride == LensSettings.OverrideModes.Physical)
                     {
                         cam.sensorSize = state.Lens.SensorSize;
                         cam.gateFit = state.Lens.GateFit;
--- a/Runtime/Core/LensSettings.cs
+++ b/Runtime/Core/LensSettings.cs
@@ -270,6 +270,7 @@ namespace Cinemachine
             FarClipPlane = farClip;
             Dutch = dutch;
             m_SensorSize = new Vector2(1, 1);
+            GateFit = Camera.GateFitMode.Horizontal;
 
 #if CINEMACHINE_HDRP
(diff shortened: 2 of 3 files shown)


Runtime/Core/LensSettings.cs
```diff
@@ -87,10 +87,10 @@ public enum OverrideModes
 
         /// <summary>
         /// This setting controls whether the Perspective/Ortho, IsPhysical, SensorSize, 
-        /// and GateFit are set in the Camera object, or are overidden here.
+        /// and GateFit are set in the Camera object, or are overridden here.
         /// </summary>
         [Tooltip("This setting controls whether the Perspective/Ortho, IsPhysical, SensorSize, "
-            + "and GateFit are set in the Camera object, or are overidden here.")]
```
glabute, line 93: thanks for catching the typo

Runtime/Behaviours/CinemachineBrain.cs
```diff
@@ -873,7 +873,7 @@ private void PushStateToUnityCamera(in CameraState state)
                     cam.lensShift = state.Lens.LensShift;
                     cam.orthographic = state.Lens.Orthographic;
                     cam.usePhysicalProperties = state.Lens.IsPhysicalCamera;
-                    if (state.Lens.IsPhysicalCamera)
+                    if (state.Lens.IsPhysicalCamera && state.Lens.ModeOverride == LensSettings.OverrideModes.Physical)
```
glabute, line 876: I don't think this is the right fix.  What if it's a physical camera with no override and the vcam wants to set the gate fit?  I think the real problem is that the gate fit (which is newly added in the vcam) doesn't get properly initialized and defaults to "None".  It's supposed to drive the camera (like the clip planes), but we need to somehow handle the case where it's uninitialized, and pull the value from the camera - otherwise we always push it

glabute (changes requested):

Runtime/Behaviours/CinemachineBrain.cs
glabute: I looked more closely, and changed my mind about this.  It seems that the GateFit is only settable n the inspector if Mode Override is physical, so this fix is good after all

glabute (approved):

PR #225, opened 2021-04-08
CMPostprocessing state bug

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-290
https://forum.unity.com/threads/depth-of-field-effect-is-broken-when-using-cinemachine-2-6-0-and-above.1088446/

Fix a regression in CMPostProcessing.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

The problem is only with Framing Transposer, because the state is incorrect in the post aim stage.
Applying PostProcessing post Finalize (or Noise) stage fixes this. However, I suspect that there was a reason for applying it post aim. If that's the case, let's discuss a solution. Note, vcam.State.FinalPosition has the correct position value.

### Package version

- [ ] Updated package version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [unreleased]
 - Switching targets (Follow, LookAt) is smooth by default. For the old behaviour, after changing the targets, set PreviousStateIsValid to false.
 - Bugfix: Reversing a blend in progress respects asymmetric blend times.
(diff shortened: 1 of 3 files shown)


gaborkb:
The problem is only with Framing Transposer, because the state is incorrect in the post aim stage.
Applying PostProcessing in Finalize state fixes this. However, I suspect that there was a reason for applying it post aim. If that's the case, let's discuss a solution. Note, vcam.State.FinalPosition has the correct position value.

glabute (changes requested):
Please add changelog entry

glabute:
Also: there is similar logic in CinemachineVolumeSettings.cs.  Do we need the same change there too?

glabute (approved):

PR #231, opened 2021-04-20
Volume layer override

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-284

CM VolumeSettings extension builds a hidden volume under the hood which arbitrarily chooses a layer from the camera's culling mask, so that it's visible.

The problem is that some users want to be able to choose which layer the volume attaches to. 

Description here: https://forum.unity.com/threads/hdrp-splitscreen-cinemachine-volume-settings-extension-issue.1043635/#post-6997481

Possible solution: an advanced setting allowing user to manually configure the layer.  By default it does what it currently does: chooses one arbitrarily

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version


gaborkb:
I am not entirely sure about the use case. I tested only that I can set a layer that will make the volume invisible.

glabute (changes requested):
Maybe we need to think about whether an advanced setting is the right solution.

Runtime/PostProcessing/CinemachineVolumeSettings.cs
```diff
@@ -99,6 +99,18 @@ public enum FocusTrackingMode
...
+        /// Global override for the layer used by CinemachineVolumes.
+        /// </summary>
+        [Tooltip("Global override for the layer used by CinemachineVolumes. ")]
+        public static LayerMask s_LayerMaskOverride;
+
+        /// <summary>
+        /// If true, enables global layer override.
+        /// </summary>
+        [Tooltip("If true, enables global layer override. ")]
+        public static bool s_LayerMaskOverrideEnabled;
```
glabute, line 112: Why static?  Having it as a global setting does not respond to the needs.  If you read the forum post, the guy wants a per-camera setting.

Runtime/PostProcessing/CinemachineVolumeSettings.cs
```diff
@@ -243,7 +257,9 @@ static void ApplyPostFX(CinemachineBrain brain)
         {
             CameraState state = brain.CurrentCameraState;
             int numBlendables = state.NumCustomBlendables;
-            var volumes = GetDynamicBrainVolumes(brain, numBlendables);
+            var currentVolume = 
+                brain.ActiveVirtualCamera.VirtualCameraGameObject.GetComponent<CinemachineVolumeSettings>();
+            var volumes = GetDynamicBrainVolumes(brain, numBlendables, currentVolume);
```
gaborkb, line 262: I dislike this part. We could discuss a better solution.

gaborkb:
Marked draft until we come back to it.

gaborkb:
See https://jira.unity3d.com/browse/CMCL-284 for more info.

PR #236, opened 2021-04-26
3rdPerson refactor

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-307
3rdPersonFollow should not set the rotation of the camera to align with CM pipeline. 3rdPersonAim can deduce rotation from Follow target's transform. This also enables other vcam types to emerge (3rdPersonFollow + any Aim component).

Relates to:
https://jira.unity3d.com/browse/CMDO-209

Tests:
Tested manually with all aim components.

1 and 2 are great rigs!

1. Aim: Same As Follow Target -> This works like 3rd Person Aim (without the hit marker)!
2. Aim: HardLookAt -> If looking at the Follow Target, then we have a simpler version of Freelook!
3. Aim: POV -> Feels weird in the example, because player rotates with POV rotations. But, otherwise it is a useful setup. Use case example: submarine periscope (POV looks around, Follow part sets itself so it is where the periscope should be). Possible with other setups too.
4. Aim: Composer -> Use case: third person follow is used like a crane on a target. Composer looks around.
Does not work well if composer is looking at Follow target. Jerky movement.

*********
TODO: changelog
Should we advertise the new interesting rigs? 

1. No hit target 3rd person camera: 3rdPersonFollow + Same As Follow Target
2. Simple Freelook: 3rdPersonFollow + HardLookAt (same target)?

Or should I create example scenes for them, and add them to this PR?

Rest of the changes don't really affect users.
*********

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version

### DIFF
--- a/Runtime/Behaviours/Cinemachine3rdPersonAim.cs
+++ b/Runtime/Behaviours/Cinemachine3rdPersonAim.cs
@@ -95,12 +95,13 @@ namespace Cinemachine
         {
             var aimDistance = AimDistance;
             var player = VirtualCamera.Follow;
-            var playerOrientation = player.transform.rotation;
-            var fwd = playerOrientation * Vector3.forward;
             
             // We don't want to hit targets behind the player
+            var fwd = Vector3.forward;
             if (player != null)
             {
+                var playerOrientation = player.transform.rotation;
+                fwd = playerOrientation * Vector3.forward;
                 var playerPos = Quaternion.Inverse(playerOrientation) * (player.position - camPos);
                 if (playerPos.z > 0)
                 {
@@ -110,8 +111,7 @@ namespace Cinemachine
             }
 
             aimDistance = Mathf.Max(1, aimDistance);
-            bool hasHit = RuntimeUtility.RaycastIgnoreTag(
-                new Ray(camPos, fwd), 
+            bool hasHit = RuntimeUtility.RaycastIgnoreTag(new Ray(camPos, fwd), 
                 out RaycastHit hitInfo, aimDistance, AimCollisionFilter, IgnoreTag);
             return hasHit ? hitInfo.point : camPos + fwd * aimDistance;
         }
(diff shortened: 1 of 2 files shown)


Runtime/Behaviours/Cinemachine3rdPersonAim.cs
```diff
@@ -90,17 +91,17 @@ void DrawReticle(CinemachineBrain brain)
...
-        Vector3 GetLookAtPoint(ref CameraState state)
+        Vector3 GetLookAtPoint(Vector3 camPos)
         {
-            var fwd = state.CorrectedOrientation * Vector3.forward;
-            var camPos = state.CorrectedPosition;
             var aimDistance = AimDistance;
-
-            // We don't want to hit targets behind the player
             var player = VirtualCamera.Follow;
+            var playerOrientation = player.transform.rotation;
```
glabute, line 98: Need null check for player

Runtime/Components/Cinemachine3rdPersonFollow.cs
```diff
@@ -206,7 +206,6 @@ void PositionCamera(ref CameraState curState, float deltaTime)
 
             // Set state
             curState.RawPosition = camPos;
-            curState.RawOrientation = targetRot;
```
glabute, line 209: I'm not so sure this is a good idea anymore.  It will break scenes that have only the 3rdPersonFollow and no 3rdPersonAim.  I think we're stuck leaving this here.  People can still add something in Aim to override this.

glabute (changes requested):

Runtime/Components/Cinemachine3rdPersonFollow.cs
gaborkb: Ok, but 3rdPersonAim should not rely on curState.RawOrientation. Or should I change that code back?

glabute (approved):

PR #241, opened 2021-05-05
Freelook prefab children are no longer visible.

gaborkb:
### Purpose of this PR

User reported some problems with freelook prefabs, when applying prefab modifications (Override -> Apply All).

1. The children can become visible. 
2. Data is lost.

I could reproduce #1, but not #2. 
I think #2 is a user error. I think this is what happens:
They modify parameter X on prefab instance A, then they modify parameter X on prefab instance B, and then they try to override all prefabs with Override -> Apply All from prefab instance B.
Instance A is not modified in this case, but this is the expected behaviour.

#1 is fixed by adding the HideChildren functionality to the prefab override callback.

https://forum.unity.com/threads/cinemachine-freelook-prefabs-seem-to-become-corrupted.532989/#post-7079488

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

Low

### Comments to reviewers

I deleted SetChildrenFlag from UpdateRigCache. The reason is that I don't see a reason for setting the flags there. However, it is possible I overlooked something, and there is a case where this is needed.

### Package version

- [ ] Updated package version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: Reversing a blend in progress respects asymmetric blend times.
 - Bugfix: CinemachineCollider's displacement damping was being calculated in world space instead of camera space.
 - Bugfix: TrackedDolly sometimes introduced spurious rotations if Default Up and no Aim behaviour.
+- Bugfix: Freelook prefabs won't get corrupted after editing the Prefab via its instances.
 
 
 ## [2.8.0-exp.1] - 2021-03-31
--- a/Runtime/Behaviours/CinemachineFreeLook.cs
+++ b/Runtime/Behaviours/CinemachineFreeLook.cs
@@ -184,6 +184,9 @@ namespace Cinemachine
             base.OnEnable();
             InvalidateRigCache();
             UpdateInputAxisProvider();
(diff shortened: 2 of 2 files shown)


glabute (changes requested):
If I disable the FreeLook component in the prefab, then drag it into the scene, the child objects appear.  
image

glabute:
Also try this:

image

then

image

gaborkb:
> If I disable the FreeLook component in the prefab, then drag it into the scene, the child objects appear.
> image

Verified.
But also found a new case, when we already have prefab instances in the scene and we disable the prefab, some instances got their children visible. Solved it by adding SetFlagsForChildren in OnValidate()
OnDisable would be the logical way for me to put SetFlagsForChildren to, but unfortunately, at that stage the flags are correctly set. The flags get unset in "deserialization" which I think happens when the prefab is modified. That's why setting in OnValidate() solves it, because it gets called when "deserialization"  happens.

It seems like after calling SetFlagsForChildren in OnValidate(), there is no need to call it in Awake(). So I removed that. Please verify.

gaborkb:
> Also try this:
> 
> image
> 
> then
> 
> image

Verified.

glabute (approved):

glabute:
Huge improvement!  Good move putting it in OnValidate

PR #257, opened 2021-05-13
CMCL-320: 3rdPersonAim extension can now specify the ray start. 

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-320

3rdPersonAim can specify ray start. Useful for setting the tip of the gun for more precise hit target calculation.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

0

### Package version

- [ ] Updated package version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [unreleased]
 - Bugfix: 3rdPersonFollow works with Aim components now. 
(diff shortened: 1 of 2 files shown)


glabute:
I'm not yet convinced that it's a good idea to add another parameter that defines the origin of the ray.  The intention is that the invisible proxy object driving the vcam should represent the firing origin and direction.  It needs to be correctly offset relative to your player.  Why not just do that?  A new transform for the gun tip just introduces extra baggage and a new way for it to be broken: what if the gun direction is not aligned with player forward?

PR #263, opened 2021-05-24
Blend condition bug

gaborkb:
### Purpose of this PR

Fix https://jira.unity3d.com/browse/CMCL-329

Blend Condition check was wrong, see my comment.

### Testing status

- [x] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version

### DIFF
--- a/Runtime/Behaviours/CinemachineBrain.cs
+++ b/Runtime/Behaviours/CinemachineBrain.cs
@@ -680,7 +680,7 @@ namespace Cinemachine
                             // Special case: if backing out of a blend-in-progress
                             // with the same blend in reverse, adjust the blend time
                             if ((frame.blend.CamA == activeCamera || 
-                                (frame.blend.CamA as BlendSourceVirtualCamera)?.Blend.CamA == activeCamera) && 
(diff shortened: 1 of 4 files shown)


Runtime/Behaviours/CinemachineBrain.cs
```diff
@@ -679,9 +679,10 @@ private void UpdateFrame0(float deltaTime)
                         {
                             // Special case: if backing out of a blend-in-progress
                             // with the same blend in reverse, adjust the blend time
-                            if (frame.blend.CamA == activeCamera
```
gaborkb, line 682: The bug is caused by this check here. When we are mid-blend, then frame.blend.CamA is not equal to activeCamera, because activeCamera is going to be a vcam, whereas frame.blend.CamA is going to be a Cinemachine.BlendSourceVirtualCamera (e.g. [Target CM Vcam] 95% from [Source CM Vcam]).

glabute (approved):

PR #284, opened 2021-06-26
Cleaned up editor tests

mjmvisser:
### Purpose of this PR

* use Assert.That instead of boolean expressions
* use parameterized tests where possible to avoid copypasta
* enclose in Tests.Editor namespace

### Testing status

- [X] Added an automated test
- [X] Passed all automated tests

### Notes to reviewers

"Why the f did you do make all of these changes"

Assert.That gives better messages when tests fail. I was also able to remove some vector comparison code that already exists in the test framework.

Parameterized tests make each test case show up individually in NUnit (and fail independently) and better shows the number of tests. At some point we'll be "graded" on how many tests we have so this is a cheap way to boost our score. It also removed a lot of code duplication.

Namespaces because good practices. Enable the "ignore whitespace" option in Github's diff to hide the noise.

### DIFF
--- a/Tests/Editor/AxisStateTests.cs
+++ b/Tests/Editor/AxisStateTests.cs
@@ -1,95 +1,90 @@
-using UnityEngine;
-using UnityEngine.TestTools;
-using NUnit.Framework;
 using System.Collections;
+using NUnit.Framework;
 using Cinemachine;
 
-public class AxisStateTests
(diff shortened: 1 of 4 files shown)


glabute (approved):
n both PlayMode and EditMode, tests are misleadingly called "CustomBlends".  Why is that?  And how to fix?

image

PR #310, opened 2021-09-02
CMCL-402 Input system should be read only once per render frame

glabute:
### Purpose of this PR

CMCL-402 Input system should be read only once per render frame
### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### DIFF
--- a/Runtime/Core/AxisState.cs
+++ b/Runtime/Core/AxisState.cs
@@ -97,7 +97,9 @@ namespace Cinemachine
         [Tooltip("Automatic recentering to at-rest position")]
         public Recentering m_Recentering;
 
-        private float mCurrentSpeed;
+        private float m_CurrentSpeed;
+        private float m_LastUpdateTime;
+        private int m_LastUpdateFrame;
 
         /// <summary>Constructor with specific values</summary>
         /// <param name="minValue"></param>
@@ -131,9 +133,11 @@ namespace Cinemachine
             m_InputAxisValue = 0;
             m_InvertInput = invert;
 
-            mCurrentSpeed = 0f;
+            m_CurrentSpeed = 0f;
             m_InputAxisProvider = null;
             m_InputAxisIndex = 0;
+            m_LastUpdateTime = 0;
+            m_LastUpdateFrame = 0;
         }
 
         /// <summary>Call from OnValidate: Make sure the fields are sensible</summary>
@@ -154,7 +158,9 @@ namespace Cinemachine
         public void Reset()
         {
             m_InputAxisValue = 0;
-            mCurrentSpeed = 0;
+            m_CurrentSpeed = 0;
+            m_LastUpdateTime = 0;
+            m_LastUpdateFrame = 0;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


mjmvisser:
Here's the test report: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/0e4f6829-3a1c-4f65-8cb5-44ff3747fadb%2FTest_Report_1%2FUtrTestReport1/TestReport.html

I'm rerunning tests on master. If they pass, then the PR has changed something about FreeLook and either indicates a problem or that the tests themselves need to be updated.

mjmvisser:
master passed: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/a84641e2-52de-4b1d-941a-1a44572df3eb%2FTest_Report_1%2FUtrTestReport1/TestReport.html
So this PR changes something that affects FreeLook. If that's intentional, you should update the tests so they pass.

gaborkb:
This fixes the multiple update problem when using fixed update.

But I still get different results depending on fps.
When I set fps to 30, the mouse input was faster than when I set it to 90 fps. I could see a difference in LateUpdate mode and FixedUpdate mode.

I tested a bit and what I see is that in SpeedMode.MaxSpeed, we call `MaxSpeedUpdate(input, deltaTime);`. In this function, at line #302, we multiply with deltaTime.
`Value += m_CurrentSpeed * deltaTime;`

If I remove this and then adjust the axis speed, then I got a frame rate independent mouse movement.

gaborkb:
> This fixes the multiple update problem when using fixed update.
> 
> But I still get different results depending on fps.
> When I set fps to 30, the mouse input was faster than when I set it to 90 fps. I could see a difference in LateUpdate mode and FixedUpdate mode.
> 
> I tested a bit and what I see is that when in SpeedMode.MaxSpeed, we call `MaxSpeedUpdate(input, deltaTime);`. In this function, at line #302, we multiply with deltaTime.
> `Value += m_CurrentSpeed * deltaTime;`
> 
> If I remove this and then adjust the axis speed, then I got a frame rate independent mouse movement.

I did some more precise tests and it seems I was wrong. The fixed update fix is enough.

gaborkb (approved):

glabute:
Still need to address the failing tests

Runtime/Core/AxisState.cs
```diff
@@ -199,6 +205,28 @@ public void SetInputAxisProvider(int axis, IInputAxisProvider provider)
...
+                m_LastUpdateTime = Time.time;
+            }
+            else
+            {
+                // CinemachineCore.UpdateVirtualCamera can change deltaTime to catch up,
+                // but that should not happen when CinemachineCore.UniformDeltaTimeOverride is set.
+                deltaTime = CinemachineCore.UniformDeltaTimeOverride; 
+                // TODO: deltaTime should not be adjusted in CinemachineCore.UpdateVirtualCamera when
+                // TODO: CinemachineCore.UniformDeltaTimeOverride is set?
+            }
```
gaborkb, line 227: Tests failed, because they use `CinemachineCore.UniformDeltaTimeOverride` (deltaTime = 0.1f). But this part could overwrite the `deltaTime` value.

--------

While looking at the code, I saw that deltaTime can also be overwritten/adjusted to catch up with frames when update method is smart (fixedSmart or smart). I think, when `UniformDeltaTimeOverride` is set, then `deltaTime` should not be adjusted. Is that true?

If yes, then we should take care of that in `CinemachineCore.UpdateVirtualCamera()`, instead of inside the `else` block here.

gaborkb: @glabute

glabute: Nice catch.  I think CinemachineCore.UpdateVirtualCamera() is ok the way it is.  UniformDeltaTimeOverride  will still be respected on a per-frame basis.

PR #316, opened 2021-09-28
CMCL-353: CmSceneTools

gaborkb:
### Purpose of this PR
See https://jira.unity3d.com/browse/CMCL-353, and its subtasks.

Add scene tools to Cinemachine using overlays.

Overlays -> Toolbar and tools on the toolbar (fov, nearfarclip, follow offset, tracked object offset).

Tools are done with Handles.

Tools can affect the following:
- vcam (fov, nearfarclip)
- transposer, orbital transposer (not when in freelook), 3rdPersonFollow (follow offset)
- framing transposer, composer, group composer, freelook (tracked object offset)

### Testing status

Manually tested all tools:

1. Does it correctly change the values it controls
2. Tool selection is exclusive with unity tools
3. Hiding the toolbar, deselects any selected cm tools
4. Only relevant tools are shown on the toolbar. Depends on what is selected in the inspector.

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

### Package version

- [ ] Updated package version


Editor/Editors/CinemachineFreeLookEditor.cs
```diff
@@ -30,6 +30,8 @@ protected override void OnEnable()
         {
             base.OnEnable();
             Target.UpdateInputAxisProvider();
+            
+            CinemachineSceneToolUtility.RegisterTool(CinemachineSceneTool.TrackedObjectOffset);
```
glabute, line 34: Why are you registering?  You are not doing anything with it.  Also, you are not handling lens tools.

glabute (changes requested):

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Cinemachine.Editor
+{
+    public enum CinemachineSceneTool
+    {
+        None,
+        FoV,
+        FarNearClip,
+        FollowOffset,
+        TrackedObjectOffset,
+    };
```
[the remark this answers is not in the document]
glabute: Why not use typeof(MyTool) instead of enum?

Editor/Editors/CinemachineVirtualCameraEditor.cs
```diff
@@ -120,6 +126,95 @@ private void OnSceneGUI()
...
             }
+
+            base.OnSceneGUI();
+            m_PipelineSet.OnSceneGUI(); // call hidden editors
+        }
+        
+        protected override void DrawSceneTools()
+        {
+            var vcam = Target;
+            if (!vcam.IsValid || vcam.m_ExcludedPropertiesInInspector.Contains("m_Lens"))
```
gaborkb, line 137: I used `vcam.m_ExcludedPropertiesInInspector.Contains("m_Lens")`, because this way there is no need to introduce a variable. What do you think?

gaborkb:
Main changes:

- More generic tool system (no enum)
- Tracked Object offset is fixed
- FOV works for physical camera
- Vcams don't know about freelook, instead they check if their lens has been hidden. I could do a variable too, but I thought this is a good solution that eliminates the need for an extra parameter.

----------------------------------------

Waiting for visual changes from Swap.

----------------------------------------

I am blocked on the following tasks:

1. Show only the label for the currently selected (or hovered tool),
2. SOLO vcam when dragging tool.

For these, I will need to change unity editor code. Because currently they do not expose the id of the handle with which I could identify which handle is used/hovered.

gaborkb:
Added a work around for the above blocked tasks:
1. Show only the label for the currently selected (or hovered tool),
2. SOLO vcam when dragging tool.

mjmvisser:
Please make sure code requiring 2021.2+ is #ifdef'd and tests pass on 2019.4+ before merging.

glabute:
Tracked Object Offset tool is still available even if vcam/FreeLook has no target.  Does this make sense?

glabute:
I'm not sure I like the automatic Solo mode.  It feels a little heavy-handed, since it wakes up damping and other time-based effects.  It's a bit shocking.  I'm leaning more towards turning this off and letting the user decide when they want to solo.  Having a Solo button in the CM toolbar strikes me as a good compromise here.

Editor/Utility/BaseEditor.cs
```diff
@@ -92,6 +92,24 @@ public override void OnInspectorGUI()
...
+        }
+
+        /// <summary>
+        /// Implement your scene handles for your components and extensions here.
+        ///
+        /// Don't forget to register/unregister your components and extensions in their
+        /// editor's OnEnable and OnDisable functions with CinemachineSceneToolUtility.RegisterTool/UnregisterTool. 
+        /// </summary>
+        /// <param name="guideLinesColor">Color of guidelines</param>
+        /// <param name="defaultColor">Color of handles, when user is not using them</param>
```
glabute, line 110: obsolete param doc

Editor/Utility/BaseEditor.cs
```diff
@@ -92,6 +92,24 @@ public override void OnInspectorGUI()
...
+        protected virtual void DrawSceneTools() {}
+
```
glabute, line 112: I think this does not belong in this class, because it's too specific to vcam inspectors, and because it will break existing derived classes thanks to the new virtual OnSceneGUI.

Instead, make a new class that inherits this and add the OnSceneGUI and DrawSceneTools methods there.  Have the tools-aware inspectors inherit from that (safe to do since they are all internal)

Editor/Utility/VcamStageEditor.cs
```diff
@@ -341,6 +354,18 @@ public void OnInspectorGUI(bool withHeader)
...
+                var ed = m_subeditors[i];
+                if (ed == null)
+                    continue;
+                
+                ed.OnSceneGUI();
+            }
+        }
+
         // Pass the dragged event down to the CM component editors
         public void OnPositionDragged(Vector3 delta)
```
glabute, line 370: This mechanism is no longer needed.  Can be removed if not an API-breaking change

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,381 @@
...
+        /// <summary>
+        /// Unregister your Type from the editor script's OnDisable function.
+        /// This way CinemachineTools will know which tools to display.
+        /// </summary>
+        /// <param name="tool">Tool to register</param>
+        public static void UnregisterTool(Type tool)
+        {
+            s_RequiredTools.Remove(tool);
+        }
+        static HashSet<Type> s_RequiredTools;
```
glabute, line 48: What if 2 components add the same tool and then one of the components is changed?  It will incorrectly unregister the tool for both.  It would be more robust to have something like ref counting

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,381 @@
...
+using UnityEngine.Experimental.Rendering;
+
+namespace Cinemachine.Editor
+{
+    /// <summary>
+    /// Static class that manages Cinemachine Tools. It knows which tool is active,
+    /// and ensures that no tools is active at the same time.
+    /// The tools and editors requiring tools register/unregister themselves here for control.
+    /// </summary>
+    static class CinemachineSceneToolUtility
```
glabute, line 16: This is effectively internal, right?  Not public?  Should not be public.

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,381 @@
...
+            
+            if (GUIUtility.hotControl == ncHandleId || GUIUtility.hotControl == fcHandleId) 
+                CinemachineBrain.SoloCamera = vcam;
+        }
+
+        public static void TrackedObjectOffsetTool(
+            CinemachineComponentBase cmComponent, ref Vector3 trackedObjectOffset, string tag)
+        {
+            var lookAtPos = cmComponent.LookAtTargetPosition;
+                var lookAtRot = cmComponent.LookAtTargetRotation;
```
glabute, line 301: some kooky indentation here

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,381 @@
...
+                    "Far Clip Plane (" + lens.FarClipPlane.ToString("F1") + ")");
+            }
+
+            
+            if (GUIUtility.hotControl == ncHandleId || GUIUtility.hotControl == fcHandleId) 
+                CinemachineBrain.SoloCamera = vcam;
+        }
+
+        public static void TrackedObjectOffsetTool(
+            CinemachineComponentBase cmComponent, ref Vector3 trackedObjectOffset, string tag)
```
glabute, line 298: You don't need the "tag" param.  You can get the tag by looking at cmComponent.Stage

glabute (changes requested):

glabute:
Try this:
1. Create a vcam, give it a follow target
2. Select the Follow Offset Tool.  Observe the tool widgets correctly drawn for the vcam
3. Change Transposer to 3rdPersonFollow
4. Tool is still selected, but widgets are not displayed until you unselect and reselect the vcam

Editor/Editors/CinemachineVirtualCameraBaseEditor.cs
```diff
@@ -356,7 +356,7 @@ protected void DrawGlobalControlsInInspector()
                     MessageType.Info);
         }
 
-        LensSettingsInspectorHelper m_LensSettingsInspectorHelper;
+        private protected LensSettingsInspectorHelper m_LensSettingsInspectorHelper;
```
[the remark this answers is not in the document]
glabute: I think I'd prefer that you just expose a "DisplayHorizontalFOV" getter that looks at the private LensSettingsInspectorHelper, rather than this kooky private protected thing

Editor/Experimental/CinemachineNewFreeLookEditor.cs
```diff
@@ -146,58 +161,59 @@ public override void OnInspectorGUI()
             DrawExtensionsWidgetInInspector();
         }
 
-        static GUIContent[] s_RigNames = 
+        internal static GUIContent[] s_RigNames =
```
glabute, line 164: keep private

Editor/Experimental/CinemachineNewFreeLookEditor.cs
```diff
@@ -146,58 +161,59 @@ public override void OnInspectorGUI()
...
         {
             new GUIContent("Top Rig"), 
             new GUIContent("Main Rig"), 
             new GUIContent("Bottom Rig")
         };
-        static int s_SelectedRig = 1;
-
-        Vector3 mPreviousPosition; // for position dragging
-        private void OnSceneGUI()
+        internal static int s_SelectedRig = 1;
```
glabute, line 170: keep private

Editor/Experimental/CinemachineNewFreeLookEditor.cs
```diff
@@ -146,58 +161,59 @@ public override void OnInspectorGUI()
...
+                {
+                    m_FovReverse = 
+                        Target.m_Lens.Orthographic ? Target.m_Lens.OrthographicSize : Target.m_Lens.FieldOfView;
                 }
+                
+                CinemachineSceneToolHelpers.FovToolHandle(newFreelook, ref newFreelook.m_Lens, 
+                    m_LensSettingsInspectorHelper == null ? false : m_LensSettingsInspectorHelper.UseHorizontalFOV, 
+                    ref m_FovReverse, ref m_SoloSetByTools);
+            }
+            else if (/*newFreelook.m_CommonLens && */CinemachineSceneToolUtility.IsToolActive(typeof(FarNearClipTool)))
```
glabute, line 206: why /*newFreelook.m_CommonLens && */ ?

Editor/Experimental/CinemachineNewFreeLookEditor.cs
```diff
@@ -146,58 +161,59 @@ public override void OnInspectorGUI()
...
+            {
+                CinemachineSceneToolHelpers.NearFarClipHandle(newFreelook, ref newFreelook.m_Lens, ref m_SoloSetByTools);
             }
-            else if (GUIUtility.hotControl == 0 && Target.UserIsDragging)
+            else if (newFreelook.Follow != null && CinemachineSceneToolUtility.IsToolActive(typeof(FollowOffsetTool)))
             {
-                // We're not dragging anything now, but we were
-                InspectorUtility.RepaintGameView();
-                Target.UserIsDragging = false;
+                CinemachineSceneToolHelpers.NewFreelookOrbitControl(newFreelook, ref m_SoloSetByTools);
```
glabute, line 212: Move NewFreelookOrbitControl logic here, avoid spaghetti

Editor/Overlays/CinemachineToolbarOverlay.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    [Overlay(typeof(SceneView), "Cinemachine")]
+    [Icon("Packages/com.unity.cinemachine/Gizmos/cm_logo.png")]
+    class CinemachineToolbarOverlay : ToolbarOverlay
+    {
+        public CinemachineToolbarOverlay()
+            : base(
+                FreelookRigSelection.id,
+#if CINEMACHINE_EXPERIMENTAL_VCAM
+                NewFreelookRigSelection.id,
```
glabute, line 23: I think this is not needed anymore

Editor/Experimental/CinemachineNewFreeLookEditor.cs
```diff
@@ -82,12 +81,28 @@ protected override void OnEnable()
...
         {
             m_PipelineSet.Shutdown();
             base.OnDisable();
+            
+#if UNITY_2021_2_OR_NEWER
+            CinemachineSceneToolUtility.UnregisterTool(typeof(SoloVcamTool));
+            CinemachineSceneToolUtility.UnregisterTool(typeof(FoVTool));
+            CinemachineSceneToolUtility.UnregisterTool(typeof(FarNearClipTool));
+            CinemachineSceneToolUtility.UnregisterTool(typeof(FollowOffsetTool));
+            CinemachineSceneToolUtility.UnregisterTool(typeof(NewFreelookRigSelection));
```
glabute, line 104: does this do anything useful?  if not, remove it

Editor/Overlays/CinemachineToolbarOverlay.cs
```diff
@@ -0,0 +1,214 @@
...
+                    InspectorUtility.RepaintGameView();
+                });
+            }
+            menu.DropDown(worldBound);
+        }
+    }
+   
+#if CINEMACHINE_EXPERIMENTAL_VCAM
+    [EditorToolbarElement(id, typeof(SceneView))]
+    class NewFreelookRigSelection : EditorToolbarDropdown
```
glabute, line 179: not needed, I think

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,564 @@
...
+        }
+        
+        static CinemachineSceneToolUtility()
+        {
+            s_ExclusiveTools = new SortedDictionary<Type, CinemachineSceneToolDelegates>(Comparer<Type>.Create(
+                (x, y) => x.ToString().CompareTo(y.ToString())));
+            s_Tools = new SortedDictionary<Type, CinemachineSceneToolDelegates>(Comparer<Type>.Create(
+                (x, y) => x.ToString().CompareTo(y.ToString())));
+            s_RequiredTools = new SortedDictionary<Type, int>(Comparer<Type>.Create(
+                (x, y) => x.ToString().CompareTo(y.ToString())));
```
glabute, line 179: Why are you using sorted dictionaries?  There will never be a large number of tools so it's probably less efficient than unsorted

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,564 @@
...
+        
+        public static void DrawLabel(Vector3 position, string text)
+        {
+            var labelOffset = HandleUtility.GetHandleSize(position) / 5f;
+            Handles.Label(position + new Vector3(0, -labelOffset, 0), text, s_LabelStyle);
+        }
+
+        static int s_ScaleSliderHash = "ScaleSliderHash".GetHashCode();
+        public static void FovToolHandle(CinemachineVirtualCameraBase vcam, ref LensSettings lens, 
+            bool isLensHorizontal, ref float fov, ref bool soloSetByTools)
```
glabute, line 268: Why are you passing in lens _and_ fov?  Isn't lens enough?  Callers have a bunch of logic to make fov from lens.  Can't that be moved here and the calling code simplified?

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,564 @@
...
+
+                if (lens.Orthographic)
+                {
+                    lens.OrthographicSize += (fov - newFov);
+                }
+                else
+                {
+                    lens.FieldOfView += (fov - newFov);
+                }
+                lens.Validate();
```
glabute, line 292: Does this get called automatically thanks to Undo?  Same question for other calls to Validate() elsewhere in this file

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,564 @@
...
+            Handles.color = followOffsetHandleIsDraggedOrHovered ? 
+                Handles.selectedColor : CinemachineSettings.CinemachineCoreSettings.ActiveGizmoColour;
+            Handles.DrawDottedLine(cmComponent.FollowTargetPosition, camPos, k_DottedLineSpacing);
+            Handles.color = originalColor;
+
+            SoloOnDrag(followOffsetHandleIsDragged, cmComponent.VirtualCamera, foHandleMaxId, ref soloSetByTools);
+        }
+        
+        public static void SoloOnDrag(
+            bool isDragged, CinemachineVirtualCameraBase vcam, int handleMaxId, ref bool soloSetByTools)
```
glabute, line 453: Why are you have a soloSetByTools param?  Can't you just have a local static and manage it yourself?  It would simplify the callers.

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,564 @@
...
+                CinemachineBrain.SoloCamera = vcam;
+            }
+            else if (soloSetByTools && handleMaxId != -1) // Handles sometimes return -1 as id, ignore those frames
+            {
+                CinemachineBrain.SoloCamera = null;
+                soloSetByTools = false;
+            }
+        }
+        
+        public static void FreelookOrbitControl(CinemachineFreeLook freelook, ref bool soloSetByTools)
```
glabute, line 468: remove this method, let the caller call OrbitControl directly.  Less spaghetti

Editor/Utility/VcamStageEditor.cs
```diff
@@ -184,6 +183,18 @@ public void OnInspectorGUI()
             DrawComponentInspector();
         }
 
+        public void OnSceneGUI()
+        {
+            if (m_ComponentEditor != null) {
```
glabute, line 188: optimization: "&& m_ComponentEditor.target != null" can be moved here

glabute: also, could cache the method for efficiency.  Reflection is expensive

Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -486,7 +486,7 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
 
         /// Serialized in order to support copy/paste
         [SerializeField][HideInInspector][NoSaveDuringPlay]
-        private CinemachineVirtualCamera[] m_Rigs = new CinemachineVirtualCamera[3];
+        internal CinemachineVirtualCamera[] m_Rigs = new CinemachineVirtualCamera[3];
```
glabute, line 489: keep private

Runtime/Components/CinemachineFramingTransposer.cs
```diff
@@ -303,7 +303,7 @@ internal Rect HardGuideRect
             }
         }
 
-        private void OnValidate()
+        internal void OnValidate()
```
glabute, line 306: I think this can stay private

glabute (changes requested):
I really like the rig editor, and the solo-on-drag.  Both feel really good to use.  Nice job.

Editor/Utility/CinemachineSceneTools.cs
gaborkb: I changed the name to _fovAfterLastToolModification_, because it was not clear at all what this fov was doing. This fov value is needed to reverse the ScaleHandles effect. This fov represents, as the name suggests now, the value of fov after the tool moved the tool.

Editor/Editors/CinemachineFramingTransposerEditor.cs
```diff
@@ -210,5 +220,65 @@ private static void DrawGroupComposerGizmos(CinemachineFramingTransposer target,
...
+                var newHandlePosition = Handles.Slider(cdHandleId, camPos, targetForward,
+                    HandleUtility.GetHandleSize(camPos) / 10f, Handles.CubeHandleCap, 0.5f);
+                if (EditorGUI.EndChangeCheck())
+                {
+                    Undo.RecordObject(framingTransposer, 
+                        "Changed FramingTransposer distance using handle in Scene View.");
+                    
+                    framingTransposer.m_CameraDistance -= 
+                        CinemachineSceneToolHelpers.SliderHandleDelta(newHandlePosition, camPos, targetForward);
+                    framingTransposer.OnValidate();
```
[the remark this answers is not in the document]
gaborkb: It is not called unfortunately. I am calling it to ensure that the lens values stay within their boundaries. E.g. if I remove this call, you can modify the lens.FieldOfView to be negative or bigger than 179.

Editor/Utility/VcamStageEditor.cs
glabute: actually, even simpler: just have VcamStageEditor cache it as a member

Editor/Utility/CinemachineSceneTools.cs
glabute: I don't see why you need the callers to manage it.  Can't you manage it locally with a static variable?

Editor/Utility/CinemachineSceneTools.cs
```diff
@@ -0,0 +1,518 @@
...
+                {
+                    lensProperty.FindPropertyRelative("OrthographicSize").floatValue += 
+                        (s_FOVAfterLastToolModification - newFov);
+                }
+                else
+                {
+                    lensProperty.FindPropertyRelative("FieldOfView").floatValue += 
+                        (s_FOVAfterLastToolModification - newFov);
+                }
+                lensProperty.serializedObject.ApplyModifiedProperties();
```
glabute, line 296: Undo isn't working for any of the lens tools.
Maybe because lensProperty.serializedObject gives you only a local copy of a lens, and not the owner object?

Editor/Utility/VcamStageEditor.cs
```diff
@@ -184,6 +183,28 @@ public void OnInspectorGUI()
             DrawComponentInspector();
         }
 
+        Dictionary<UnityEditor.Editor, System.Reflection.MethodInfo> m_OnSceneGUIs =
```
glabute, line 186: Why a dictionary?  Just need a member `MethodInfo m_OnSceneGUIMethod`

glabute (changes requested):
Need to remove leftover instances of `public void OnVcamPositionDragged(Vector3 delta)` (there are 4 of them)

Editor/Utility/CinemachineSceneTools.cs
gaborkb: Hmm... it works for me.

PR #322, opened 2021-10-13
CMCL-570: Refactored menu based creation to be consistent with the rest of Unity

MechWarrior99:
### Purpose of this PR
The current implementation used to create Cinemachine Cameras in the Hierarchy is inconsistent with the rest of Unity. This leads to Cinemachine feeling lower quality and some things to not produce the expected result.
Main changes
- Names of created objects are now consistent with the rest of Unity. Example being "Vcam" is now "Virtual Camera".
- Duplicate names of created objects is now consistent with the rest of Unity. Example being "Vcam1" is now "Virtual Camera (1)". This will also support the "Game Object Naming" scheme setting option that is present in newer versions of Unity.
- It is now easier to add more options to the menu later with less boilerplate and places for failure.

### Testing status
Manually used each menu item and ensured that the resulting GameObjects and components are the same as the current implementation.
Each menu item was manually tested with both undo and redo.

- [x] Manually tested 

### Documentation status

### Technical risk
There were definitions for versions < 2019.3 which were removed as the minimum Unity version for the package is 2019.4. This should not cause any problems but worth mentioning.

~`CinemachineEditorAnalytics.SendCreateEvent(string name)` is now sent for each object that is created instead of per menu item. The name that is sent is also the same as the non-unique name of the object that is created. 
This results in slightly less specific names for the events, and a slightly higher number. I don't suspect that this will be a problem, but is a significant enough change that it is worth mentioning.~

### Comments to reviewers
Currently names of created GameObjects are the same as the menu items, but perhaps they would be better being more explicit? For example "Mixing Virtual Camera" instead of "Mixing Camera", however doing so does make the names relatively longer.
~The "2D Camera" item also creates a GameObject called "2D Camera", I'm not sure if it would be better to just call it "Virtual Camera" instead like it was.~

### DIFF
--- a/Editor/Menus/CinemachineMenu.cs
+++ b/Editor/Menus/CinemachineMenu.cs
@@ -147,7 +147,7 @@ namespace Cinemachine.Editor
         [MenuItem(m_CinemachineGameObjectRootMenu + "2D Camera", false, m_GameObjectMenuPriority)]
         static void Create2DCamera(MenuCommand command)
         {
-            CinemachineEditorAnalytics.SendCreateEvent("Virtual Camera");
+            CinemachineEditorAnalytics.SendCreateEvent("2D Camera");
             var vcam = CreateCinemachineObject<CinemachineVirtualCamera>(
                 "Virtual Camera", command.context as GameObject, true);
             vcam.m_Lens = MatchSceneViewCamera(vcam.transform);
--- a/Editor/Menus/CinemachineMenu.cs
+++ b/Editor/Menus/CinemachineMenu.cs
@@ -77,6 +77,7 @@ namespace Cinemachine.Editor
             CreateDefaultVirtualCamera(parentObject: stateDrivenCamera.gameObject);
         }
 
+#if CINEMACHINE_PHYSICS
         [MenuItem(m_CinemachineGameObjectRootMenu + "ClearShot Camera", false, m_GameObjectMenuPriority)]
         static void CreateClearShotVirtualCamera(MenuCommand command)
         {
@@ -86,10 +87,9 @@ namespace Cinemachine.Editor
 
             // We give the camera a child as an example setup
             var childVcam = CreateDefaultVirtualCamera(parentObject: clearShotCamera.gameObject);
-#if CINEMACHINE_PHYSICS
             Undo.AddComponent<CinemachineCollider>(childVcam.gameObject).m_AvoidObstacles = false;
-#endif
         }
+#endif
 
         [MenuItem(m_CinemachineGameObjectRootMenu + "Dolly Camera with Track", false, m_GameObjectMenuPriority)]
         static void CreateDollyCameraWithPath(MenuCommand command)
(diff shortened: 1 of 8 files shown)


Editor/Utility/InspectorUtility.cs
```diff
@@ -143,18 +143,6 @@ public static string NicifyClassName(string name)
...
-        // Temporarily here
-        /// <summary>
-        /// Create a game object.  Uses ObjectFactory if the Unity version is sufficient.
-        /// </summary>
-        /// <param name="name">Name to give the object</param>
-        /// <param name="types">Optional components to add</param>
-        /// <returns></returns>
-        public static GameObject CreateGameObject(string name, params Type[] types)
-        {
-            return ObjectFactory.CreateGameObject(name, types);
```
glabute, line 155: You can't delete this because it's a public API.  Instead, deprecate it.

glabute:
Thanks for the PR.  Nice work.
Probably the change in analytics will have to be reverted.
Notes on style for future reference:
- Unity is moving towards a "leave out what is redundant" style, which implies that labeling fields and methods "private" is undesirable, since they are that way by default
- Lines more than 100 characters or so wide should be split.  Sidescrolling is evil.

Editor/Menus/CinemachineMenu.cs
```diff
@@ -31,327 +28,262 @@ static void CreateFixedSignalDefinition()
...
-                + "Unity 2018.4 is unable to auto-detect the presence of PostProcessing, so you must "
-                + "manually add a define to your player settings to enable the code.\n\n"
-                + "To enable support for PostProcessing v2, please do the following:\n\n"
-                + "1. Delete the CinemachinePostProcessing folder from your assets, if it's present\n\n"
-                + "2. Open the Player Settings tab in Project Settings\n\n"
-                + "3. Add this define: CINEMACHINE_POST_PROCESSING_V2";
-
-            EditorUtility.DisplayDialog("Cinemachine Adapter Code for PostProcessing V2", message, "OK");
-        }
+            CinemachineEditorAnalytics.SendCreateEvent("Virtual Camera");
```
gaborkb, line 150: This changed from 2D camera to Virtual Camera.

Editor/Timeline/CinemachineShotEditor.cs
```diff
@@ -51,9 +51,10 @@ static SyncCacheEnabledSetting()
         }
 #endif
 
-        static public CinemachineVirtualCameraBase CreateStaticVcamFromSceneView()
-        {
-            CinemachineVirtualCameraBase vcam = CinemachineMenu.CreateStaticVirtualCamera();
+        static public CinemachineVirtualCameraBase CreatePassiveVcamFromSceneView()
+       {
+            CinemachineEditorAnalytics.SendCreateEvent("Virtual Camera");
```
gaborkb, line 56: This is new analytics. The Create Event analytics is specifically for the Create Menu. I don't think this is wanted.

gaborkb (changes requested):

gaborkb (approved):

PR #328, opened 2021-10-21
CMCL-353: No Undo, instead using serializedObject

gaborkb:
I am not sure if this is better. In some ways, it is, in some it is not.

What do you think?
I would need to do the same for FoV, NearClipFarClip, and TrackedObjectOffset. 
I would need to pass lens as a SerializedProperty for the first two, and trackedObjectOffset as a SerializedProperty for the last.

These:
Undo.RecordObject(vcam, "Changed clip plane using handle in scene view.");
Undo.RecordObject(vcam, "Changed FOV using handle in scene view.");
Undo.RecordObject(cmComponent, "Change Tracked Object Offset using handle in Scene View.");

### DIFF
--- a/Editor/Editors/CinemachineFreeLookEditor.cs
+++ b/Editor/Editors/CinemachineFreeLookEditor.cs
@@ -150,8 +150,12 @@ namespace Cinemachine
             }
             else if (freelook.Follow != null && CinemachineSceneToolUtility.IsToolActive(typeof(FollowOffsetTool)))
             {
-                CinemachineSceneToolHelpers.OrbitControlHandle(freelook, 
-                    ref freelook.m_Orbits, ref s_SelectedRig);
+                var so = new SerializedObject(freelook);
+                var orbits = so.FindProperty(() => freelook.m_Orbits);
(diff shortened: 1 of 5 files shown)


Editor/Editors/CinemachineFreeLookEditor.cs
```diff
@@ -150,8 +150,12 @@ void DrawSceneTools()
             }
             else if (freelook.Follow != null && CinemachineSceneToolUtility.IsToolActive(typeof(FollowOffsetTool)))
             {
-                CinemachineSceneToolHelpers.OrbitControlHandle(freelook, 
-                    ref freelook.m_Orbits, ref s_SelectedRig);
+                var so = new SerializedObject(freelook);
+                var orbits = so.FindProperty(() => freelook.m_Orbits);
+                if (CinemachineSceneToolHelpers.OrbitControlHandle(freelook, orbits, ref s_SelectedRig))
+                {
+                    so.ApplyModifiedProperties();
```
glabute, line 157: OrbitControlHandle can do ApplyModifiedProperties so it doesn't have to return a bool

glabute (changes requested):

PR #336, opened 2021-10-27
CMCL-339: Near Clip Plane negative value not kept if override not set to Orthographic

gaborkb:
### Purpose of this PR

Fix for https://jira.unity3d.com/browse/CMCL-339

Fix:
- Validate near clip plane when pulling values from camera.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - New feature: Added scene view overlay tools for Cinemachine components.
 - Regression fix: Lookahead works again.
 - Cinemachine3rdPersonAim exposes AimTarget, which is the position of where the player would hit.
+- Bugfix: Negative Near Clip Plane value is kept when camera is orthographic.
 
 
 ## [2.8.0] - 2021-07-13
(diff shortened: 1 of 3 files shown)


gaborkb:
I think a better alternative would be something like this, but it is an API change (or a new API):
`public void Validate(CinemachineBrain brain)
        {
            if (brain != null)
                SnapshotCameraReadOnlyProperties(brain.OutputCamera);
            if (!Orthographic)
                NearClipPlane = Mathf.Max(NearClipPlane, 0.001f);
            // ...`

glabute (changes requested):
I don't like this fix.  Already it is very confusing with the flags and the overrides, and this makes it worse.  
What about removing the NearClipPlane check from OnValidate?  The check could be put in the property drawer instead.  Then you wouldn't need the flag.

gaborkb:
> I don't like this fix. Already it is very confusing with the flags and the overrides, and this makes it worse. What about removing the NearClipPlane check from OnValidate? The check could be put in the property drawer instead. Then you wouldn't need the flag.

Me neither. :) I wanted to talk to you before review, I added the flag just to highlight what's wrong. I like your alternative :)

glabute (commented):
Code is good, but

glabute (changes requested):
Code is good but missing changelog and PR description is stale

glabute (approved):

PR #347, opened 2021-11-08
CMCL-517: Support for input system in the samples

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-517

When Input System is added to Cinemachine, the Cinemachine examples are throwing a lot of errors, because they are trying to read from the old backends (UnityEngine.Input). 

There is a guard that checks if the old backends are disabled (ENABLE_LEGACY_INPUT_MANAGER). I added this to all example scripts that use Input, and a Debug.Log that tells users how they can fix it.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Samples status

### Comments to reviewers
I added a guard to AxisState too, but I think it is better if that error is actually displayed as it was before. See my review comment for more details.

### Package version

- [ ] Updated package version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: Negative Near Clip Plane value is kept when camera is orthographic.
 - Regression fix: could not change the projection of the main camera if a CM virtual camera is active.
 - Regression fix: Axis input was ignoring CM's IgnoreTimeScale setting.
-- Bugfix: Cinemachine assigns a default input controller delegate that returns 0 when Input System is present.
+- Bugfix: Cinemachine assigns a default input controller delegate that returns 0 when the legacy input system is disabled.
 - Cinemachine example scenes show informative text when used with Input System instead of throwing error messages.
 
 ## [2.9.0-pre.1] - 2021-10-26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: Negative Near Clip Plane value is kept when camera is orthographic.
 - Regression fix: could not change the projection of the main camera if a CM virtual camera is active.
 - Regression fix: Axis input was ignoring CM's IgnoreTimeScale setting.
-- 
+- Bugfix: Cinemachine assigns a default input controller delegate that returns 0 when Input System is present.
+- Cinemachine example scenes show informative text when used with Input System instead of throwing error messages.
 
 ## [2.9.0-pre.1] - 2021-10-26
 - Added ability to directly set the active blend in CinemachineBrain.
--- a/Runtime/Core/AxisState.cs
+++ b/Runtime/Core/AxisState.cs
@@ -219,11 +219,13 @@ namespace Cinemachine
             
             if (m_InputAxisProvider != null)
                 m_InputAxisValue = m_InputAxisProvider.GetAxisValue(m_InputAxisIndex);
+#if ENABLE_LEGACY_INPUT_MANAGER
             else if (!string.IsNullOrEmpty(m_InputAxisName))
             {
                 try { m_InputAxisValue = CinemachineCore.GetInputAxis(m_InputAxisName); }
                 catch (ArgumentException e) { Debug.LogError(e.ToString()); }
             }
+#endif
 
             float input = m_InputAxisValue;
             if (m_InvertInput)
--- a/Runtime/Core/AxisState.cs
+++ b/Runtime/Core/AxisState.cs
@@ -219,13 +219,11 @@ namespace Cinemachine
             
             if (m_InputAxisProvider != null)
                 m_InputAxisValue = m_InputAxisProvider.GetAxisValue(m_InputAxisIndex);
-#if ENABLE_LEGACY_INPUT_MANAGER
             else if (!string.IsNullOrEmpty(m_InputAxisName))
             {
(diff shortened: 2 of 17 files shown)


Runtime/Core/AxisState.cs
```diff
@@ -219,11 +219,13 @@ public bool Update(float deltaTime)
             
             if (m_InputAxisProvider != null)
                 m_InputAxisValue = m_InputAxisProvider.GetAxisValue(m_InputAxisIndex);
+#if ENABLE_LEGACY_INPUT_MANAGER
             else if (!string.IsNullOrEmpty(m_InputAxisName))
             {
                 try { m_InputAxisValue = CinemachineCore.GetInputAxis(m_InputAxisName); }
                 catch (ArgumentException e) { Debug.LogError(e.ToString()); }
             }
+#endif
```
gaborkb, line 228: This might not be desired, because it hides the fact that there is no correct input provider assigned!

gaborkb: @glabute

Samples~/Cinemachine Example Scenes/Shared/Scripts/InputSystemHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+        public static void EnableBackendsWarningMessage()
+        {
+            if (Time.realtimeSinceStartup - s_LastMessageTime > 5)
+            {
+                Debug.Log(
+                    "Old input backends are disabled. Cinemachine examples use UnityEngine.Input. " +
+                    "To enable input in our examples: Edit > Project Settings > Player, " +
+                    "set Active Input Handling to Both!\n" +
+                    "Note: This is probably because the Input System Package has been added to the project. To use " +
+                    "Cinemachine Virtual Cameras with Input System Package, add CinemachineInputProvider to them.");
```
gaborkb, line 18: @JWalker0007

AntoineCharton (approved):
Awesome to add support for it and guidelines! Only thing check for the warning message the standardized name of the Built in or Classic input system.
https://docs.unity3d.com/Manual/com.unity.inputsystem.html

Samples~/Cinemachine Example Scenes/Shared/Scripts/InputSystemHelper.cs
JWalker0007, line 18: with the Input System Package......

JWalker0007 (approved):

glabute (approved):

glabute (changes requested):
Wait, I take back my approval.  See AxisState.cs

Runtime/Core/AxisState.cs
glabute: No, wait.  This change should be reverted, and instead change the default implementation of CinemachineCore.GetInputAxis

Runtime/Core/CinemachineCore.cs
```diff
@@ -72,7 +72,11 @@ public static CinemachineCore Instance
         /// <summary>Delegate for overriding Unity's default input system.
         /// If you set this, then your delegate will be called instead of
         /// System.Input.GetAxis(axisName) whenever in-game user input is needed.</summary>
+#if ENABLE_LEGACY_INPUT_MANAGER
         public static AxisInputDelegate GetInputAxis = UnityEngine.Input.GetAxis;
+#else
+        public static AxisInputDelegate GetInputAxis = null;
```
glabute, line 78: What happens if it gets called?  No one is checking for null.  I think it should be an empty function instead.

glabute (changes requested):

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bugfix: Negative Near Clip Plane value is kept when camera is orthographic.
 - Regression fix: could not change the projection of the main camera if a CM virtual camera is active.
 - Regression fix: Axis input was ignoring CM's IgnoreTimeScale setting.
-- 
+- Bugfix: Cinemachine assigns a default input controller delegate that returns 0 when Input System is present.
```
glabute, line 11: I think you mean " when the legacy input system is disabled"

Runtime/Core/CinemachineCore.cs
```diff
@@ -72,7 +72,11 @@ public static CinemachineCore Instance
         /// <summary>Delegate for overriding Unity's default input system.
         /// If you set this, then your delegate will be called instead of
         /// System.Input.GetAxis(axisName) whenever in-game user input is needed.</summary>
+#if ENABLE_LEGACY_INPUT_MANAGER
         public static AxisInputDelegate GetInputAxis = UnityEngine.Input.GetAxis;
+#else
+        public static AxisInputDelegate GetInputAxis = delegate { return 0; };
```
glabute, line 78: Did you test this?  I am unfamiliar with this syntax.

glabute (approved):

Runtime/Core/CinemachineCore.cs
gaborkb: Yes, I tested it.
What did you have in mind instead?

gaborkb: I could also do this:
`public static AxisInputDelegate GetInputAxis = axisName => 0;`

But to me the delegate version was clearer.

gaborkb: The OrbTransposer in CameraPositionTest tests this indirectly, when the legacy input system is disabled.

glabute: No worries.  Like I said, I was unfamiliar with the syntax so wanted to make sure it was tested.

PR #350, opened 2021-11-12
CMCL-570: Refactored menu based creation to be consistent with the rest of Unity

gaborkb:
### Purpose of this PR

This branch is an intermediate layer between our master and the kind outside user (MechWarrior99) who made some improvements to our Menu. 
Original PR: https://github.com/Unity-Technologies/com.unity.cinemachine/pull/322

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


gaborkb:
Changelog is missing. Maybe we should acknowledge MechWarrior99? @mjmvisser @bradweiers ?

glabute:
I don't think changelog is necessary.  There is no visible change for the user.  Also, I would not credit the contributor in this case.

glabute (approved):

MechWarrior99:
> I don't think changelog is necessary. There is no visible change for the user. Also, I would not credit the contributor in this case.

I don't mind not being credited as others had to 'finish' it. But I would argue that this *is* a visible change for the user. New Cinemachine Cameras will now have a different name, this will be more apparent to users who have other Cinemachine cameras using the default names, or if they create cameras frequently.

PR #356, opened 2021-11-19
CMCL-644: Freelook recentering is affected by Time Scale even with Ignore Time Scale on

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-644
The problem came from inconsistencies and delta time in user input handling. Now delta time is always real delta time for input.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


Runtime/Core/AxisState.cs
```diff
@@ -210,11 +210,7 @@ public bool Update(float deltaTime)
                 return false;
             m_LastUpdateFrame = Time.frameCount;
 
-            // Cheating: we want the render frame time, not the fixed frame time
```
glabute, line 213: you need to leave this comment, or something like it. What's going on isn't obvious because the caller is passing in deltaTime

Runtime/Core/AxisState.cs
```diff
@@ -210,11 +210,7 @@ public bool Update(float deltaTime)
...
-            if (CinemachineCore.UniformDeltaTimeOverride >= 0)
-                deltaTime = CinemachineCore.UniformDeltaTimeOverride; 
-            else if (Time.inFixedTimeStep && deltaTime >= 0 && m_LastUpdateTime != 0)
```
glabute, line 216: you lost the important check for `if (deltaTime >= 0 && m_LastUpdateTime != 0)`

Runtime/Core/AxisState.cs
```diff
@@ -424,6 +423,9 @@ public void RecenterNow()
             /// <param name="recenterTarget">The value that is considered to be centered</param>
             public void DoRecentering(ref AxisState axis, float deltaTime, float recenterTarget)
             {
+                deltaTime = Time.realtimeSinceStartup - m_LastUpdateTime;
```
glabute, line 426: you need to comment why we're setting deltaTime even though the user passed it in.  Also, you need to add some checks (same as above).  You can't just blindly set this.  deltaTime < 0 has special meaning in CM.

glabute (changes requested):

glabute (approved):

gaborkb:
Freelook tests are no longer deterministic, because it uses deltaTime (before it was using `deltaTime = CinemachineCore.UniformDeltaTimeOverride`).
Should we remove them?

glabute (approved):

PR #371, opened 2021-12-09
CMCL-711: These tests fail on yamato, we don't know why.

gaborkb:
### Purpose of this PR

Why do these test fail? https://jira.unity3d.com/browse/CMCL-711
1. We need to WaitForFixedUpdate to ensure that physics is updated (e.g. when we need to raycast).
2. Tests were not deterministic due to frameDelta time compensation. Sometimes, some tests need to wait a couple frames to ensure physics (fixed update) is updated. This would causes a couple frame compensation in the test giving different results, sometimes so much that the test would fail.

https://jira.unity3d.com/browse/CMCL-941
Internal option to disable frame delta time to have deterministic test results.

### Testing status

- [x] Added an automated test
- [ ] Manually tested 

### Documentation status

- [ ] Updated [CHANGELOG](https://keepachangelog.com/en/1.0.0/), No need, because this does not add anything public. It adds unit tests, and enables us to do tests deterministically. 

### Package version

- [ ] Updated package version


Runtime/Core/CinemachineCore.cs
```diff
@@ -346,6 +346,8 @@ UpdateTracker.UpdateClock updateClock
             int frameDelta = (updateClock == UpdateTracker.UpdateClock.Late)
                 ? Time.frameCount - status.lastUpdateFrame
                 : s_FixedFrameCount - status.lastUpdateFixedFrame;
+            
+            // TODO: The frameDelta causes instabilities in our tests. Should this be used when manual update is used?
```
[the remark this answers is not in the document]
gaborkb: I think when using ManualUpdate, we should not automatically compensate with frameDelta, because the user should be in control. If they are using ManualUpdate, and want to compensate, they can supply a compensated delta time. (compensation for lost/not updated frames).

Runtime/Behaviours/CinemachineCollider.cs
```diff
@@ -223,11 +223,10 @@ public float ApplyDistanceSmoothing(float distance, float smoothingTime)
             }
             public void UpdateDistanceSmoothing(float distance)
             {
-                float now = CinemachineCore.CurrentTime;
-                if (m_SmoothedDistance == 0 || distance <= m_SmoothedDistance)
+                if (m_SmoothedDistance == 0 || distance < m_SmoothedDistance)
                 {
                     m_SmoothedDistance = distance;
-                    m_SmoothedTime = now;
+                    m_SmoothedTime = CinemachineCore.CurrentTime;
```
glabute, line 229: good catch

Runtime/Core/CinemachineCore.cs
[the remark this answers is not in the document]
glabute: That's possibly true, but I don't know how we would implement it, because this method cannot know about the brain update mode (there might be multiple brains in the scene.  Which brain?)

Perhaps we could do something like: if _any_ brain has manual update mode, it sets something in Core which causes `deltaTime *= frameDelta` to not happen. 

Still, this won't give the desired user control, because the compensation would have to be per-vcam, not global.  What value would they choose?

Runtime/Behaviours/CinemachineCollider.cs
```diff
@@ -326,6 +328,7 @@ public override float GetMaxDampTime()
                     // Apply damping
                     if (deltaTime >= 0 && VirtualCamera.PreviousStateIsValid)
                     {
+                        // TODO: when delta time is 0, the displacement here is wrong Damper.Damp should return -extra.previousDisplacement
```
glabute, line 331: Can you explain?  To my thinking, if deltaTime is 0 then displacement should not change - i.e. it should be the same as previousDisplacement.

Runtime/Behaviours/CinemachineCollider.cs
```diff
@@ -326,6 +328,7 @@ public override float GetMaxDampTime()
...
                         displacement = extra.previousDisplacement + Damper.Damp(
```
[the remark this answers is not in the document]
glabute: No it shouldn't.  Damper.Damp is supposed to return the amount of the value that gets applied during the given deltaTime.  If deltaTime is 0, then this amount should be 0.  That's what Damper.Damp does.

Tests/Runtime/CmColliderTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            UnityEngine.Object.Destroy(obstacle);
+            var previousPosition = m_Vcam.State.FinalPosition;
+            
+            yield return IncreaseTimeAndWaitForOnePhysicsFrame(m_DeltaTime);
+            m_Brain.ManualUpdate();
+            
+            Assert.That(previousPosition, !Is.EqualTo(m_Vcam.State.FinalPosition).Using(Vector3EqualityComparer.Instance));
+            
+            // Because of frame delta in CinemachineCore, this cannot be checked!
+            // Assert.That(originalCamPosition, !Is.EqualTo(m_Vcam.State.FinalPosition).Using(Vector3EqualityComparer.Instance));
```
glabute, line 148: A minimal hack fix would be to add an internal bool to CinemachineCore to turn off frame delta compensation

glabute (changes requested):

Tests/Runtime/CmColliderTests.cs
gaborkb: That would work. To have an option to turn it off. Then our tests will be stable, and we can have deltaTimeOverride = 0.1 working deterministicly.

Runtime/Behaviours/CinemachineCollider.cs
gaborkb: You are right. I meant for my test to work. It would need to return -extra.previousDisplacement.

glabute, line 331: Why is the comment still here?

glabute (approved):
Please remove the incorrect comment

PR #375, opened 2021-12-15
CMCL-87: Splines and Cinemachine

gaborkb:
### Purpose of this PR

This branch covers most of this epic: https://jira.unity3d.com/browse/CMCL-87

It deprecates old cinemachine internal splines (currently only when Spline package is added), and adds new components that use Spline package directly. 

Splines is not a hard dependency currently, but I think it should be from CM 3.0. 
In this version: 

- if the Splines package is present in the project, then user can use the new component (SplineDolly) and old deprecated components;
- if the Splines package is not present in the project, then user can only use deprecated components.

Samples are also updated with a separate SplineSample package -> SplineDollyCart, and SplineDolly examples.

TODO: Spline guards are setup to check for Splines version 0.9.9. This should be changed to 1.0.0 after review. Current Spline package master branch is 1.0.0-pre something.

### Testing status

- [x] Added an automated test
- [x] Manually tested 

### Documentation status

### Comments to reviewers
Officially supported unity version for the Spline package is 2022.1+. It also works in 2021, but Splines don't work in 2020 and 2019.
Use Spline package master branch: https://github.com/Unity-Technologies/com.unity.splines. 

------------------------

There are couple of bugs in the Spline package. If you find any problems let me know. I'll check it. It may be a bug that has already been reported or a new one (in this case, I'll check if it is my bug or theirs - and let them know if theirs).

Bugs that may trip you up:
- SplineData (override value: Roll, Speed, Offset) in the Inspector may not be visible. They may throw errors. They may collapse.
- Spline may not update its internal states if you change its values in the Inspector. Use the overlay tool in the Scene view (wiggle it, if you want your changes in the inspector to apply).
- This is fixed on latest master now:
Lerp is wrong at the Start and End point of the spline. This affects overrides (Roll, Speed, Offset). This may give strange result at those points. 
See in the below gif, the carts get stuck for a little on the closer side, but are smooth on the farther side.
Lerp is not smooth

### Package version

- [ ] Updated package version


Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+    /// <summary>
+    /// A Cinemachine Virtual Camera Body component that constrains camera motion
+    /// to a Spline.  The camera can move along the path.
+    ///
+    /// This behaviour can operate in two modes: manual positioning, and Auto-Dolly positioning.
+    /// In Manual mode, the camera's position is specified by animating the Path Position field.
+    /// In Auto-Dolly mode, the Path Position field is animated automatically every frame by finding
+    /// the position on the path that's closest to the virtual camera's Follow target.
+    /// </summary>
+    [DocumentationSorting(DocumentationSortingAttribute.Level.UserRef)]
```
glabute, line 18: This is an obsolete attribute.  Don't use it in new classes.  Remove everywhere in this file.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+        /// <summary>How to set the virtual camera's Up vector.  This will affect the screen composition.</summary>
+        [Tooltip("How to set the virtual camera's Up vector.  This will affect the screen composition, because "
+            + "the camera Aim behaviours will always try to respect the Up direction.")]
+        public CameraUpMode m_CameraUp = CameraUpMode.Default;
+
+        /// <summary>
+        /// How aggressively the camera tries to track the target rotation (x: Pitch, y: Yaw, z: Roll).
+        /// Smaller numbers are more responsive. Larger numbers give a more heavy, slowly responding camera.</summary>
+        /// </summary>
+        public Vector3 m_RotationDamping = Vector3.zero;
```
glabute, line 94: Would it make more sense for this to be a single float value, used to damp the slerp from old rotation to the new one?  Why do we need component-wise damping here?  See CinemachineTransposer,cs line 267 for implementation example.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+            var a = Mathf.Max(m_Damping.x, Mathf.Max(m_Damping.y, m_Damping.z)); 
+            var b = Mathf.Max(d2.x, Mathf.Max(d2.y, d2.z)); 
+            return Mathf.Max(a, b); 
+        }
+        
+        /// <summary>
+        /// Subscribe to onSplineChanged if you'd like to react to changes to the Spline attached to this vcam.
+        /// This action is invoked by the Spline's changed event when a spline property is modified.
+        /// </summary>
+        public event Action onSplineChanged;
```
glabute, line 175: Is this following a correct naming convention?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+        { 
+            var d2 = AngularDamping;
+            var a = Mathf.Max(m_Damping.x, Mathf.Max(m_Damping.y, m_Damping.z)); 
+            var b = Mathf.Max(d2.x, Mathf.Max(d2.y, d2.z)); 
+            return Mathf.Max(a, b); 
+        }
+        
+        /// <summary>
+        /// Subscribe to onSplineChanged if you'd like to react to changes to the Spline attached to this vcam.
+        /// This action is invoked by the Spline's changed event when a spline property is modified.
```
glabute, line 173: Is this editor-only or does it also work in builds?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+            else if (m_Track.TryGetComponent(out m_RollOverrideExtension)) {}
+        }
+
+        CinemachineSplineRollOverrideExtension m_RollOverrideExtension; // don't use this directly
+        CinemachineSplineRollOverrideExtension RollOverrideExtension
+        {
+            get
+            {
+                if(m_RollOverrideExtension == null)
+                    UpdateSplineRollOverrideCache();
```
glabute, line 222: This doesn't feel very robust.  What if user adds a rollover extension then replaces it with another?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+                return;
+            }
+            if (pathSpline.Count == 1)
+            {
+                curState.RawPosition = m_PreviousCameraPosition = pathSpline[0].Position;
+                curState.RawOrientation = m_PreviousOrientation = pathSpline[0].Rotation;
+                return;
+            }
+            
+            m_PathPosition =
```
glabute, line 251: style comment: the code is correct but it would feel better to me if you did not modify m_PathPosition here.  Why not instead just declare var newPathPosition = m_PathPosition here, and then always modify newPathPosition?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+            // splines work with normalized position by default, so we convert m_PathPosition to normalized at the start
+            var pathSpline = m_Track.Spline;
+            if (pathSpline == null || pathSpline.Count == 0)
+            {
+                return;
+            }
+            if (pathSpline.Count == 1)
+            {
+                curState.RawPosition = m_PreviousCameraPosition = pathSpline[0].Position;
+                curState.RawOrientation = m_PreviousOrientation = pathSpline[0].Rotation;
```
glabute, line 247: what about path offset?  Come to think of it, do you even need a special case here?  Won't the regular stream just gracefully handle a degenerate spline with only one knot?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+        /// <param name="deltaTime">Used for damping.  If less that 0, no damping is done.</param>
+        public override void MutateCameraState(ref CameraState curState, float deltaTime)
+        {
+            if (!IsValid)
+            {
+                m_PathPosition = 0;
+                return;
+            }
+
+            // splines work with normalized position by default, so we convert m_PathPosition to normalized at the start
```
glabute, line 238: this comment more properly belongs a few lines lower

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+            var newPathOrientation = 
+                Vector3.SqrMagnitude(localTangent) == 0 || Vector3.SqrMagnitude(localUp) == 0 ? 
+                    Quaternion.identity : Quaternion.LookRotation(localTangent, localUp);
+            
+            if (RollOverrideExtension != null && RollOverrideExtension.enabled)
+            {
+                float roll = RollOverrideExtension.RollOverride.Evaluate(pathSpline, newPathPosition, 
+                    PathIndexUnit.Normalized, new UnityEngine.Splines.Interpolators.LerpFloat());
+                var rollRotation = Quaternion.AngleAxis(-roll, localTangent);
+                newPathOrientation = Quaternion.LookRotation(localTangent, rollRotation * localUp);
```
glabute, line 309: This will fail with obscure console message if tangent is zero.  This happens at start and end points of the path in your sample scene.  Need better handling of this case.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+                offset = Damper.Damp(offset, m_Damping.z, deltaTime);
+                newPathPosition = m_PreviousNormalizedPathPosition - offset;
+            }
+            m_PreviousNormalizedPathPosition = newPathPosition;
+            m_Track.Evaluate(newPathPosition, 
+                out var localPosition, out var localTangent, out var localUp);
+            Vector3 newCameraPos = localPosition;
+            var newPathOrientation = 
+                Vector3.SqrMagnitude(localTangent) == 0 || Vector3.SqrMagnitude(localUp) == 0 ? 
+                    Quaternion.identity : Quaternion.LookRotation(localTangent, localUp);
```
glabute, line 302: Is it possible that localTangent and localUp are parallel?  What happens in that case?  Try making a path that goes straight up and adding roll to it.  Does it work as expected?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+            newCameraPos += m_PathOffset.y * offsetY;
+            newCameraPos += m_PathOffset.z * offsetZ;
+
+            // Apply damping to the remaining directions
+            if (deltaTime >= 0 && VirtualCamera.PreviousStateIsValid)
+            {
+                Vector3 currentCameraPos = m_PreviousCameraPosition;
+                Vector3 delta = (currentCameraPos - newCameraPos);
+                Vector3 delta1 = Vector3.Dot(delta, offsetY) * offsetY;
+                Vector3 delta0 = delta - delta1;
```
glabute, line 326: are you erroneously combining deltaX and deltaZ here?  You already damped deltaZ, I think

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,399 @@
...
+        {
+            get
+            {
+                switch (m_CameraUp)
+                {
+                    case CameraUpMode.PathNoRoll:
+                    case CameraUpMode.FollowTargetNoRoll:
+                        return new Vector3(m_RotationDamping.x, m_RotationDamping.y, 0);
+                    case CameraUpMode.Default:
+                        return Vector3.zero;
```
glabute, line 387: This feels wrong.  No rotation damping if default up mode?  Why?

Runtime/Helpers/CinemachineSplineRollOverrideExtension.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// that has SplineDolly as Body component.
+    /// - When CinemachineSplineRollExtension is added to a gameObject that has SplineContainer, then any vcam that reads
+    /// that SplineContainer will be affected by the CinemachineSplineRollExtension overrides. It works as a global
+    /// override in this case.
+    /// - When CinemachineSplineRollExtension is added to a vcam, then the overrides are local, they only affect that
+    /// one vcam.
+    /// </summary>
+    [ExecuteInEditMode]
+    [DisallowMultipleComponent]
+    public class CinemachineSplineRollOverrideExtension : MonoBehaviour
```
glabute, line 19: Why is it called RollOverride?  Why override?

Runtime/Helpers/CinemachineSplineRollOverrideExtension.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        /// <summary>
+        /// Roll (in angles) along the forward direction for specific location on the track.
+        /// When placed on a SplineContainer, this is going to be a global override that affects all vcams using the Spline.
+        /// Whem placed on a vcam, this is going to be a local override that only affects that vcam.
+        /// </summary>
+        [Tooltip("Roll (in angles) along the forward direction for specific location on the track.\n" +
```
glabute, line 26: we roll _around_ a direction, not _along_ a direction

Runtime/Helpers/CinemachineSplineRollOverrideExtension.cs
```diff
@@ -0,0 +1,67 @@
...
+#if UNITY_EDITOR
+                !TryGetComponent(out splineContainer) && // this is needed by the RollHandle to work in the scene view
+#endif
+                TryGetComponent(out CinemachineVirtualCamera vcam))
+            {
+                // set splineContainer through the vcam
+                var body = vcam.GetCinemachineComponent(CinemachineCore.Stage.Body);
+                if (body != null)
+                {
+                    var splineDataCacheUpdate = body.GetType().GetMethod("UpdateSplineRollOverrideCache",
```
glabute, line 53: You can't use reflection at runtime - not even once.  Apparently it negatively impacts performance from then on.

Runtime/Helpers/CinemachineSplineRollOverrideExtension.cs
```diff
@@ -0,0 +1,67 @@
...
+                    {
+                        splineDataCacheUpdate.Invoke(body, null);
+                    }
+                }
+            }
+        }
+    }
+    
+    [AttributeUsage(AttributeTargets.Field)]
+    public class RollHandleAttribute : SplineDataHandleAttribute {}
```
glabute, line 65: Should be called _SplineRollHandleAttribute_

glabute (changes requested):
Did not look at the samples yet

glabute:
Why is the offset interpolation weird?
weirdinterpolation

gaborkb:
> Why is the offset interpolation weird? weirdinterpolation
Good question. I forwarded it to the Splines team.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: Yes. Spline team is also doing the same convention. Their changed action: m_TrackCache.Spline.changed.
I've setup my IDE with unity's code convention template, so it should let me know when I am not following convention.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: I copied it from Tracked dolly, but I agree that just one rotation damping is enough. We discussed it with Swap too.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: Good question. I assumed it was not editor only, because the spline docs did not say it.  I checked and it is editor only. So I updated our docs, and pushed a PR to spline team too :)

Runtime/Helpers/CinemachineSplineRollOverrideExtension.cs
gaborkb: I did not know that! Thanks for pointing it out! I searched around and found that it makes gc calls slower :S

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: It works, because I get notified from by the rollOverrideExtension.

I need to rethink this though 👍

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: Currently, SplineUtility does not handle cases when spline only has one knot. When Spline team fixes this, then we can remove this special case.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: I'll look into in depth. 
I used the code from TrackedDolly. I assumed it was correct.

gaborkb: Yes, that's the case. deltaX and deltaZ are damping the same thing.

gaborkb: But, z Damping only damps on the path, because we use the Z damped value to get the position on the spline. Then we apply damping X Y to this value. So I think this is correct.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+            {
+                // convert follow target into spline local space, because SplineUtility works in spline local space
+                SplineUtility.GetNearestPoint(spline, 
+                    m_Spline.transform.InverseTransformPoint(FollowTargetPosition), out _, out normalizedSplinePosition, 
+                    m_AutoDolly.m_SearchResolution, m_AutoDolly.m_SearchIteration);
+                m_SplinePosition = SplineUtility.ConvertIndexUnit(spline, normalizedSplinePosition, 
+                    PathIndexUnit.Normalized, m_PositionUnits);
+                
+                // Apply the spline position offset
+                normalizedSplinePosition += m_AutoDolly.m_PositionOffset;
```
glabute, line 255: Is m_PositionOffset always in normalized units?  It should be using the same units as m_SplinePosition

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+            // Set the orientation and up
+            Quaternion newOrientation = GetCameraOrientationAtSplinePoint(newSplineOrientation, curState.ReferenceUp);
+            if (deltaTime >= 0 && VirtualCamera.PreviousStateIsValid)
+            {
+                Vector3 relative = (Quaternion.Inverse(m_PreviousOrientation)
+                    * newOrientation).eulerAngles;
+                for (int i = 0; i < 3; ++i)
+                    if (relative[i] > 180)
+                        relative[i] -= 360;
+                relative = Damper.Damp(relative, m_AngularDamping, deltaTime);
```
glabute, line 333: I think you need to do a proper spherical damping here, not a componentwise euler damping.
See implementation in CinemachineTransposer line 270

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+            }
+            m_PreviousNormalizedSplinePosition = normalizedSplinePosition;
+            m_Spline.Evaluate(normalizedSplinePosition, 
+                out var localPosition, out var localTangent, out var localUp);
+            Vector3 newCameraPos = localPosition;
+            if (Vector3.SqrMagnitude(localTangent) != 0)
+            {
+                m_CorrectedTangent = localTangent;
+            }
+            if (Vector3.SqrMagnitude(localUp) != 0)
```
glabute, line 288: is it ok to do hard equalities in these kinds of tests?  Is `m_Spline.Evaluate` expected to output unit vectors or zero?  If so, then document it and do a hard check for zero instead of computing magnitudes for nothing.

Also: are these expected to be either both zero or both nonzero?  If it can be that only one of them is zero then it's possible for them to become parallel in some cases, which is a problem.  This needs to be clarified.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+            var newSplineOrientation = Quaternion.LookRotation(m_CorrectedTangent, m_CorrectedUp);
+
+            var rollExt = rollExtension;
+            if (rollExt != null && rollExt.enabled)
+            {
+                float roll = rollExt.RollOverride.Evaluate(spline, normalizedSplinePosition, 
+                    PathIndexUnit.Normalized, new UnityEngine.Splines.Interpolators.LerpFloat());
+                var rollRotation = Quaternion.AngleAxis(-roll, m_CorrectedTangent);
+                newSplineOrientation = Quaternion.LookRotation(m_CorrectedTangent, rollRotation * m_CorrectedUp);
+            }
```
glabute, line 301: This is only used in some cameraUp modes, yet you are calculating it all the time.  Maybe it would make sense to move this logic into `GetCameraOrientationAtSplinePoint`, which knows whether or not it's needed.

Runtime/Helpers/CinemachineSplineRollExtension.cs
```diff
@@ -0,0 +1,71 @@
...
+                if (body != null)
+                {
+                    var splineDolly = body as CinemachineSplineDolly;
+                    if (splineDolly != null)
+                    {
+                        splineDolly.rollExtension = this;
+#if UNITY_EDITOR
+                        splineContainer = splineDolly.m_Spline; // this is needed by the RollHandle to work in the scene view
+#endif
+                    }
```
glabute, line 60: See my other comment.  I don't think this class should know about virtual cameras.  Instead, expose an API that vcams (or other objects) can use.  Think of this class as nothing but a passive data container (like a DOTS component).

Runtime/Helpers/CinemachineSplineRollExtension.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// that has SplineDolly as Body component.
+    /// - When CinemachineSplineRollExtension is added to a gameObject that has SplineContainer, then any vcam that reads
+    /// that SplineContainer will be affected by the CinemachineSplineRollExtension overrides. It works as a global
+    /// override in this case.
+    /// - When CinemachineSplineRollExtension is added to a vcam, then the overrides are local, they only affect that
+    /// one vcam.
+    /// </summary>
+    [ExecuteInEditMode]
+    [DisallowMultipleComponent]
+    public class CinemachineSplineRollExtension : MonoBehaviour
```
glabute, line 19: Can you just call this `CinemachineSplineRoll`?  I got confused with the current name and thought that it was a CinemachineExtension when it isn't.

glabute (changes requested):
Also: the SplineDolly sample doesn't make sense.  Is the help text inaccurate?  Why is there a 1-knot spline?

glabute:
I have not looked properly at the samples yet.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: I am not sure if that's right. 

newSplineOrientation orientation affects the position offset of the camera (offsetX, Y, Z). I took TrackedDolly as a base, and there 
`var newSplineOrientation = Quaternion.LookRotation(m_CorrectedTangent, m_CorrectedUp);` affects the offset, then GetCameraOrientationAtSplinePoint only affects `curState.RawOrientation`.

It is possible that is not correct though.

Maybe GetCameraOrientationAtSplinePoint should affect the position offset?

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: You are right, that's better to check for 0,0,0.

Tangent is not unit vector, but it return 0,0,0, when tangent is set to be 0,0,0. Otherwise it returns a correct tangent vector.
UpVector is unit, but it cannot be 0,0,0.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+
+            var spline = m_Spline.Spline;
+            if (spline == null || spline.Count == 0)
+            {
+                return;
+            }
+            if (spline.Count == 1)
+            {
+                curState.RawPosition = m_PreviousCameraPosition = spline[0].Position;
+                curState.RawOrientation = m_PreviousOrientation = spline[0].Rotation;
```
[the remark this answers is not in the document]
gaborkb: Currently, when spline has only 1 knot, the evaluate functions don't work (they throw an error). When they fix this, I can remove this special check.

But you are right, this is not correct. :)

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,371 @@
...
+            private get
+            {
+                if (m_RollExtension == null)
+                    m_Spline.TryGetComponent(out m_RollExtension); // check if our spline has roll extension
+
+                return m_RollExtension;
+            }
+            set
+            {
+                m_RollExtension = value; // called by CinemachineSplineRollExtension
```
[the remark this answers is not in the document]
gaborkb: I'll rethink it!

Runtime/Helpers/CinemachineSplineRollExtension.cs
gaborkb: I'll rethink it!

gaborkb:
> I thought that the naming standards indicate that property get/set should be PascalCase, not camelCase. Am I wrong?

I am totally confused on this. Do I follow Cinemachine legacy naming or the unity convention?

glabute:
> I am totally confused on this. Do I follow Cinemachine legacy naming or unity convention?
New Unity convention seems to be PascalCase for all public APIs.  We should use it.
Exception: to be consistent with legacy CM code, we prepend m_ to our member variables (even though we hate it).

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,368 @@
...
+            m_AngularDamping = Mathf.Clamp(m_AngularDamping, 0, 20);
+        }
+
+        CinemachineSplineRoll m_Roll; // don't use this directly
+        CinemachineSplineRoll SplineRoll
+        {
+            get
+            {
+                if (m_Roll == null) {
+                    transform.parent.TryGetComponent(out m_Roll); // check if vcam has CinemachineSplineRoll
```
glabute, line 200: This is not correct.  You are assuming stuff about the vcam's structure.  This will throw an exception if you use it with NewVcam.  Use CinemachineComponentBase.VirtualCamera to get the vcam owner object.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,368 @@
...
+            }
+
+            m_Damping.x = Mathf.Clamp(m_Damping.x, 0, 20);
+            m_Damping.y = Mathf.Clamp(m_Damping.y, 0, 20);
+            m_Damping.z = Mathf.Clamp(m_Damping.z, 0, 20);
+            m_AngularDamping = Mathf.Clamp(m_AngularDamping, 0, 20);
+        }
+
+        CinemachineSplineRoll m_Roll; // don't use this directly
+        CinemachineSplineRoll SplineRoll
```
glabute, line 195: If no spline roll is present, this will keep trying to fetch it every frame from 2 different sources.  Is this a performance issue?  How expensive is TryGetComponent?

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,368 @@
...
+                    PathIndexUnit.Normalized, m_PositionUnits);
+                
+                // Apply the spline position offset
+                normalizedSplinePosition += SplineUtility.ConvertIndexUnit(
+                    spline, m_AutoDolly.m_PositionOffset, m_PositionUnits, PathIndexUnit.Normalized);
+            }
+
+            if (deltaTime >= 0 && VirtualCamera.PreviousStateIsValid)
+            {
+                float maxUnit = 1; // we are always using normalized unit [0-1]
```
glabute, line 257: just hardcode it instead of making a variable, or use const

Runtime/Helpers/CinemachineSplineRoll.cs
```diff
@@ -0,0 +1,45 @@
+#if CINEMACHINE_UNITY_SPLINES
```
glabute, line 1: This file should be in Behaviours directory, hot Helpers

glabute (changes requested):

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,368 @@
...
+                m_CorrectedTangent = localTangent;
+                if (m_CorrectedTangent.Equals(localUp))
+                {
+                    // Find tangent perpendicular to localUp, localUp is always normalized
+                    var newTangent = new Vector3(localUp.z, localUp.z, -localUp.x - localUp.y);
+                    m_CorrectedTangent = !newTangent.Equals(Vector3.zero)
+                        ? newTangent
+                        : new Vector3(-localUp.y - localUp.z, localUp.x, localUp.x);
+                }
+            }
```
glabute, line 293: Can you explain to me what this code is doing?  I think the comment is out of date.  It looks to me like you're going to get a pop when it executes.  Not that it ever will because the Equals test is too strong for Vector3

glabute:
There seems to be some weirdness still in the spline stuff.  If I add a knot to the spline, the distance evaluation does not always recognize it.  Sometimes switching the spline EditModeType back and forth will fix it.  I haven't identified a pattern. 

Have you noticed anything like that?

glabute:
When the spline is looped, it should be possible to continually increase or decrease the path position and it should wrap.  Right now it gets clamped at the loop point, which is wrong.

glabute:
I find the roll editing to be very difficult.  The single line visualization of the spline makes it hard to understand what the path is doing.  I think it might make sense to put back the railroad tracks.

glabute:
Why does this say percentage when the value is 0...1?  

image

Editor/Editors/CinemachineTrackedDollyEditor.cs
```diff
@@ -1,11 +1,14 @@
...
 using UnityEditor;
 using UnityEngine;
-
 namespace Cinemachine.Editor
 {
     [CustomEditor(typeof(CinemachineTrackedDolly))]
     [CanEditMultipleObjects]
+#if UNITY_2022_1_OR_NEWER
+    [Obsolete("CinemachineTrackedDolly has been deprecated. Add Splines package to your project, and use CinemachineSplineDolly instead.", false)]
+#endif
```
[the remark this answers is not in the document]
gaborkb: The reason why I added it, is because otherwise Unity is throwing warnings in the console complaining the CmTrackedDolly is deprecated. I could either add pragma suppression for each usage in this file or this one #if obsolete tag #endif.

gaborkb:
> Why does this say percentage when the value is 0...1?
> 
> image

I had the same question. Spline returns the value like that. It's like that in their samples too. I'll check if they are going to fix it. I think they meant normalized.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: The tangent is always non-zero, expect when it is set to be 0 by the user. That's why I checked for exact equality.

gaborkb:
> There seems to be some weirdness still in the spline stuff. If I add a knot to the spline, the distance evaluation does not always recognize it. Sometimes switching the spline EditModeType back and forth will fix it. I haven't identified a pattern.
> 
> Have you noticed anything like that?

Yes. I had a lot of problems with that. I reported it to them. I was told to use the spline handle instead of the editor in the meantime. But it was supposed to be fixed by now.
I could reproduce it on latest master. I had to wiggle the spline handle to make the internal state update.

Runtime/Components/CinemachineSplineDolly.cs
gaborkb: Yes, that's not so good.
I checked in the profiler and it takes 0.02 for this call, which seems a lot. If I'd just return null, then it takes 0.00.

A potential solution would be:
- CmSplineRoll knows about vcams. Ping vcam, when CmSplineRoll is added to a vcam.
- CmSplineRoll knows about SplineContainer. When CmSplineRoll is added to a SplineContainer, it triggers an onSplineChanged, to which SplineDolly is registered. When this is called, we update m_Roll.

This way, we won't need to poll.

We had something like this in a previous iteration, but we did not like it.

glabute: Why don't you just cache it in Awake?  Only attempt a TryGetComponent if editor and not playing.

Runtime/Components/CinemachineSplineDolly.cs
glabute: it's actually this equals that I was referring to: `if (m_CorrectedTangent.Equals(localUp))`.
In any case I reimplemented this in my visualization branch, but there are still issues

Runtime/Behaviours/CinemachineSplineRoll.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// When placed on a vcam, this is going to be a local override that only affects that vcam.
+        /// </summary>
+        [Tooltip("Roll (in angles) around the forward direction for specific location on the track.\n" +
+            "- When placed on a SplineContainer, this is going to be a global override that affects all vcams using the Spline.\n" +
+            "- When placed on a vcam, this is going to be a local override that only affects that vcam.")]
+        [SplineRollHandle]
+        public SplineData<float> RollOverride;
+
+#if UNITY_EDITOR
+        internal SplineContainer SplineContainer; // SplineRollHandle needs this for drawing the handles
```
glabute, line 31: This seems to be unused.  Why not just get rid of it?

Editor/Editors/CinemachineSplineDollyEditor.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            DrawRemainingPropertiesInInspector();
+        }
+
+        [DrawGizmo(GizmoType.Active | GizmoType.NotInSelectionHierarchy
+                | GizmoType.InSelectionHierarchy | GizmoType.Pickable, typeof(CinemachineSplineDolly))]
+        static void DrawGizmos(CinemachineSplineDolly dolly, GizmoType selectionType)
+        {
+            if (Selection.activeGameObject == dolly.VirtualCamera.gameObject)
+                DrawDollyGizmo(dolly, Color.green, 1); // GML todo: make this configurable
```
glabute, line 37: Do you have any thoughts on how we should deal with this hardcoding?

Editor/Editors/CinemachineSplineDollyEditor.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        static void DrawDollyGizmo(CinemachineSplineDolly dolly, Color pathColor, float width)
+        {
+            var spline = dolly.m_Spline;
+            if (spline == null || spline.Spline == null || spline.Spline.Count == 0)
+                return;
+
+            var length = spline.CalculateLength();
+            var numSteps = Mathf.RoundToInt(Mathf.Clamp(length / width, 3, 1000));
```
glabute, line 47: this 1000 also is a little ad-hoc.  Not sure of the best way to make this make sense.  The idea is to be able to draw very long paths without killing performance.

Editor/Editors/CinemachineSplineDollyEditor.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            // Draw the path
+            Color colorOld = Gizmos.color;
+            Gizmos.color = pathColor;
+
+            var mesh = new Mesh();
+            mesh.SetVertices(vertices);
+            mesh.SetNormals(normals);
+            mesh.SetIndices(indices, MeshTopology.Lines, 0);
+            Gizmos.DrawWireMesh(mesh);
```
glabute, line 90: Drawing this complex thing as a single mesh is hugely better performing than doinbg a large number of DrawLine calls.  This is a technique that we should consider bringing to any other heavy line-based gizmos we might have.

Editor/Editors/CinemachineTrackedDollyEditor.cs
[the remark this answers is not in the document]
glabute: ok

Editor/Menus/CinemachineMenu.cs
```diff
@@ -97,14 +100,23 @@ static void CreateClearShotVirtualCamera(MenuCommand command)
...
+#else
+            var path = CreateCinemachineObject<CinemachineSmoothPath>(
+                "Dolly Track", command.context as GameObject, false);
+#pragma warning disable 618 // disable obsolete warning
             AddCinemachineComponent<CinemachineTrackedDolly>(vcam).m_Path = path;
+#pragma warning restore 618
+#endif
         }
 
         [MenuItem(m_CinemachineGameObjectRootMenu + "Dolly Track with Cart", false, m_GameObjectMenuPriority)]
```
glabute, line 122: This menu entry still gives the old-fashioned path.  What should we do about this?

Runtime/Behaviours/CinemachineSplineRoll.cs
glabute: Wait, I take that back.  We're going to need it for proper implementation of the railroad track

Editor/Menus/CinemachineMenu.cs
gaborkb: We could remove it, when splines are installed. It is discoverable in our examples. In CM 3, we could remove it completely.

Editor/Editors/CinemachineSplineDollyEditor.cs
gaborkb: We could add a color to Cinemachine preferences or use Boundary Object color.

gaborkb: Screen Shot 2022-02-16 at 2 06 16 PM

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,494 @@
...
+
+        /// <summary>
+        /// Adds spline to SplineDolly.
+        /// </summary>
+        /// <param name="spline">The spline to add. If null, then any spline on SplineDolly will be removed.</param>
+        public void AddSplineContainer(SplineContainer spline)
+        {
+            m_Spline = spline;
+            UpdateSplineData();
+        }
```
gaborkb, line 178: I needed to add this to avoid recomputing spline length and knot count. This might not be needed if Splines will have a built in input sanitizer.

Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -0,0 +1,494 @@
...
+
+            var spline = m_Spline.Spline;
+            if (spline == null || spline.Count == 0)
+            {
+                return;
+            }
+
+#if false
+            m_CameraPosition = spline.SanitizeInterpolationValue(m_CameraPosition, m_PositionUnits);
+#else
```
gaborkb, line 204: This is the API that spline team might add. Might be different, I just did a PR to their repo.

gaborkb:
Closing, because this is part of cm3.

PR #390, opened 2022-01-28
CMCL-769: Add a Timeline-controlled Camera

krisrok:
### Purpose of this PR

Adds a specialised `CinemachineTimelineCamera` that can be used with Timeline in conjunction with `CinemachineCameraTrack`. It works exactly like the existing `CinemachineTrack` (now `CinemachineBrainTrack`) but instead of writing directly to the Brain, the `CinemachineTimelineCamera` acts like any other virtual camera and can be used as such, e.g. can be mixed with other cameras.

This way, nesting and reusable Timelines are now easily possible.

### Testing status

Manually tested inside the two added sample scenes as well as in my more complex testing setup.

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

I've added no documentation yet with exception of comments and the sample scenes.

### Technical risk

I'd say risk is pretty low in terms of existing features, but here are some pointers:
- Introduced two abstract base classes to prevent code-duplication: CinemachineTrackBase and CinemachineMixerBase. 
- Renamed `CinemachineTrack` to `CinemachineBrainTrack` (same GUID, serialization should be fine)
- Renamed `CinemachineMixer` to `CinemachineBrainMixer` (no serialization, should be fine too)
- All other functionality is additional.

### Comments to reviewers

Please have a look at the sample scenes.

As stated in the `CinemachineTimelineCamera` class [comment](https://github.com/krisrok/com.unity.cinemachine/blob/fbf30b88261894e803a6d123c32987a77f4d1d0d/Runtime/Timeline/CinemachineTimelineCamera.cs#L25), the camera stays uninitialized until there is some input coming in from a `CinemachineCameraTrack` (or `CinemachineCameraMixer` to be exact). Also, the camera's state stays unchanged after the track's playback ends.
We could introduce some dependency on the `PlayableDirector` and force-initialize the state. Or make the before/after behaviour configurable. But I wanted to keep this simple for now.

I've had some trouble with the extensions first, they were processed addtively with every update. So e.g. the offset of `CinemachineCameraOffset` was added every frame.
I guessed this is because the extensions alter the state by reference, so now I keep a copy of the last known state and pass this into the `InvokePostPipelineStageCallback` call. See [here](https://github.com/krisrok/com.unity.cinemachine/blob/fbf30b88261894e803a6d123c32987a77f4d1d0d/Runtime/Timeline/CinemachineTimelineCamera.cs#L165). Hope this is viable.

I've also considered to build upon the `ICameraOverrideStack` interface introduced [here](https://github.com/Unity-Technologies/com.unity.cinemachine/commit/195dbb809c08c29a293c865fe74156b473dde397) but decided against it because we do not need multiple override sources for this.
On the other hand, it'd be possible for multiple tracks to use the same camera as target so it might not be the worst idea to use the same interface. This should be very easy to do.


krisrok:
I've noticed jittery camera movement due to the internal state updates not working correctly in a setup using a TrackedDolly with Auto-Dolly enabled.

We have two ways to avoid it, both are documented in the TimelineCamera sample scene with 92dbf44:
A) Move all sub-cameras to be child GameObjects of the `CinemachineTimelineCamera` so the `IsLiveChild()` method works as intended. Same as for e.g. `CinemachineMixingCamera`. This way, `CinemachineBrain` will update them correctly.
B) Set the sub-camera's `Standby Update` to `Never` so they are NOT updated by the `CinemachineBrain`. Instead, the `CinemachineTimelineCamera` handles the update.

In 8190e69 now there is a check if the sub-camera is nested. We update it manually only if it is not.
For `StandbyUpdateMode.RoundRobin` and `StandbyUpdateMode.Always` this results in doubled updates which messes with e.g. damping. In those cases we can avoid jitter by passing -1 for `deltaTime` but then damping will be skipped.

glabute:
Hello krisrok,
This PR is interesting and I'd like to take a look at it.  For legal reasons I can't until you sign that contributor's agreement.
Thanks for your contribution, we really appreciate it.

krisrok:
Hi @glabute, thanks for taking a look!
I've already signed the CLA and reissued the PR after recommitting with the corrections needed for the agreement stuff.
For me it reads now: `license/cla — Contributor License Agreement is signed.` Is there anything else I should do?

glabute:
Oh that's good.  My mistake for not seeing it.

glabute:
This is an interesting idea, but a correct implementation is tricky.

I think it would be appropriate for multiple CameraTracks acting on the same TimelineCamera to behave consistently with multiple BrainTracks acting on the same Brain.  In other words, the full `ICameraOverrideStack` interface should be implemented.

For me, having 2 kinds of tracks is confusing because it's not obvious which one to use.  I think many people would choose Camera Track when what they really want is a Brain Track.  I'm not sure of the best way to address this.  It might be interesting to create a new base class that implements `ICameraOverrideStack` and derive both Brain and TimelineCamera from that.  That way, you wouldn't need a new kind of track.

In the current implementation, as a user I'm a little confused about what vcams are permitted to be added to the Camera Track.  Is is only vcam children, or can we add any vcam?  If it's only the children, then it's consistent with other "manager-style" vcams (like MixingCamera and ClearShot).  Also, if that gets enforced, then it becomes impossible to add the TimelineCamera itself to the track - which you can now do and cause stack overflows.

In the current implementation, TimelineCamera is calling `InternalUpdateCameraState` for its clips - which is breaking the paradigm used elsewhere in CM and also is the cause problems relating to cameras being updated multiple times per frame - which must never happen because it messes with noise and damping.  If you look at the MixingCamera implementation, you will see that it does not call `InternalUpdateCameraState` for its children, but allows `CinemachineCore` to manage that.  You should follow that pattern.  `CinemachineCore` will guarantee that a vcam is only ever updated once per frame.  No need for m_LastState.

krisrok:
Wow, thanks for all your input. There are some different views and hope we can work something out together!

> I think it would be appropriate for multiple CameraTracks acting on the same TimelineCamera to behave consistently with multiple BrainTracks acting on the same Brain. In other words, the full `ICameraOverrideStack` interface should be implemented.

Yep, agreed.

> For me, having 2 kinds of tracks is confusing because it's not obvious which one to use. I think many people would choose Camera Track when what they really want is a Brain Track. I'm not sure of the best way to address this.

Having only one track would be nice and clean but here are my thoughts against it from the top of my head:
- Having two separate tracks this way it's clear they have different implications. Sure, they both blend cameras but there are still nuances.
- It'd suggest interchangability but putting a TimelineCamera into the track slot where the Brain was before it would not "just work" anymore (especially if you want to restrict it to vcam children as stated below).

Maybe it's "just" a naming problem? I agree it's a valid point "Camera Track" sounds more like what you'd want to use. We're operating on cameras after all.

> It might be interesting to create a new base class that implements `ICameraOverrideStack` and derive both Brain and TimelineCamera from that. That way, you wouldn't need a new kind of track.

This would surely work -- but shoving a parent class in there just to make it work with Unity's serialization is not the right way in my opinion :/

Using the interface directly would be best but the Timeline editor and Inspector cannot handle changing the track's binding like this: `[TrackBindingType(typeof(ICameraOverrideStack), TrackBindingFlags.None)]`, unfortunately.
Maybe there is a way to write editor code for the track so we can customize the binding property field?
 
> In the current implementation, as a user I'm a little confused about what vcams are permitted to be added to the Camera Track. Is it only vcam children, or can we add any vcam? If it's only the children, then it's consistent with other "manager-style" vcams (like MixingCamera and ClearShot).

Yes, it'd be consistent with other manager cams. But I really do not like this limitation at all, also not on e.g. the MixingCamera. I think it renders the whole system pretty inflexible. I'd rather take a look at the `IsLiveChild()` system and for alternatives that do _not_ need the child cameras to be actual children in the hierarchy to be honest.

Without going into much detail: In my primary use case for adopting CM we have a modular system made of prefabs you can use to put together a working scene. Cameras are also part of the prefabs to be able to have different views of the content. I want to be able to blend between those cameras dynamically but also to be able to author them e.g. in some kind of predefined Timeline. Even though cameras might be scattered all over the hierarchy.

> Also, if that gets enforced, then it becomes impossible to add the TimelineCamera itself to the track - which you can now do and cause stack overflows.

Good catch! A simple self-referencing could be checked for easily even without the parent-child-restriction. But what about nested Timelines and TimelineCameras? I'm not able to think this through right now.
 
> In the current implementation, TimelineCamera is calling `InternalUpdateCameraState` for its clips - which is breaking the paradigm used elsewhere in CM and also is the cause problems relating to cameras being updated multiple times per frame - which must never happen because it messes with noise and damping. If you look at the MixingCamera implementation, you will see that it does not call `InternalUpdateCameraState` for its children, but allows `CinemachineCore` to manage that. You should follow that pattern. `CinemachineCore` will guarantee that a vcam is only ever updated once per frame. No need for `m_LastState`.

I agree the `CinemachineCore` should manage all those updates. With the last two commits I've addressed this problem in part. Only updating the child cams when they're not a "LiveChild" of the TimelineCamera, see [here](https://github.com/Unity-Technologies/com.unity.cinemachine/pull/390/commits/8190e694196b97439584de0d69badec6fd1d2526).
This is no real solution though of course. It is this way because I cling to the idea of "more flexible managers" I've tried to express above :)

PR #403, opened 2022-02-15
add visualization - wip needs to be configurable

glabute:
Add railroad-track visualization to CM SplineDolly.
It's a crude first stab - colors and sizes need to be not hardcoded.
I think it helps with the roll visualization.
It also helps with testing.


gaborkb (approved):
This is great! Added some minor fixes in the editor!

PR #406, opened 2022-02-16
Dev/spline visualization2

glabute:
Move railroad tracks to be with the SplineRoll component, so that they are visible also when roll is on spline container.
Optimize the RollCache.
Fix SplineContainer cache so that it continues to work if roll component gets moved.


gaborkb (approved):
I pushed a commit, because the dolly track was not fully drawn when the spline was looped. Also, now all buffers are allocated the exact size (indices had an extra iteration allocated (6)).

PR #425, opened 2022-04-07
CMCL-514: Hovering and knowing which handles are picked/dragged

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-514
We checked for ids with a hack. Now the ids can be tracked correctly for each handle (unity version 2022.2.0a9 and above), so we use that instead.

******************
EDIT: EXPLANATION
Handles have an id with which we can check if the handle is picked or hovered.

The "hack" version was needed, because there was no exposed API for some of the handles to get the handle's id, or to set the handle's id on creation. The "hack" version checked the last used id with this call: `GUIUtility.GetControlID(FocusType.Passive)`, before creating the handle, and after creating the handle. This way, I knew that the created handle should be between beforeID and afterID. The reason why I needed a range, because handles are created on threads (so I had cases when beforeID + 1 was not my handle).
This could (with low probability) cause a small bug, that you pick a non-cm handle, but our cm-handle code would think you pick it. Because this non-cm handle was unluckily created right after my beforeID call or right before my afterID call.

The new version uses the new API for the handles that did not have expose an ID field on creation. So before creating the handle, we create a set of ids, and we pass it to the handle creation method. This way, we know exactly which ids the created handle has.
******************

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status
**No user facing change, so I did not add docs.**

- [N/A] Updated [CHANGELOG](https://keepachangelog.com/en/1.0.0/)
- [N/A] Updated README (if applicable)
- [N/A] Commented all public classes, properties, and methods
- [N/A] Updated user documentation

### Package version

- [ ] Updated package version

### DIFF
--- a/com.unity.cinemachine/Editor/Editors/Cinemachine3rdPersonFollowEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/Cinemachine3rdPersonFollowEditor.cs
@@ -73,9 +73,14 @@ namespace Cinemachine.Editor
 
                 EditorGUI.BeginChangeCheck();
                 // shoulder handle
-                var sHandleMinId = GUIUtility.GetControlID(FocusType.Passive); // TODO: KGB workaround until id is exposed
+#if UNITY_2022_2_OR_NEWER
+                var sHandleIds = Handles.PositionHandleIds.@default;
+                var newShoulderPosition = Handles.PositionHandle(sHandleIds, shoulderPosition, heading);
+#else
+                var sHandleMinId = GUIUtility.GetControlID(FocusType.Passive);
                 var newShoulderPosition = Handles.PositionHandle(shoulderPosition, heading);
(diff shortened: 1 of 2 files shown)


glabute (commented):
How exactly am I supposed to test this?  The descriptions are unclear.

gaborkb:
You are right, it is very vague and unclear. I editted the description (EDIT: EXPLANATION). Sorry about that! 👍

glabute:
How do I test it?

gaborkb:
> How do I test it?

I tested it by using the handles in the scene view. Checking each handle whether they highlight when I hover over/pick it.

glabute:
Does not work with 2022.0a8.  What version of 2022 is required?  PR should specify.

gaborkb:
It should work 2022.2.0a9 and above

PR #430, opened 2022-04-14
Dev/cinemachine 3.0 unittests

gaborkb:
### Purpose of this PR

Unit tests for valid CmCamera cache, that run in both PlayMode and EditMode. We can have tests that run in both if they are placed at the Root tests folder (namespace Tests).

Found a bug with the unit tests in CmCamera.cs. Fixed it.
The bug was caused in `UpdatePipelineCache` when destroying components. The components invalidate `m_Pipeline` by setting it to null in their OnDisable call. But after destroying the component, we are accessing the array.

### Testing status

- [x] Added an automated test
- [x] Manually tested


glabute:
I had a few issues with this.
1. the implementation of UpdatePipelineCache() should not create or destroy components.  It's just about the cache.
2. removing that invalidated one of the tests - there is no guarantee when adding multiple components in the same stage which one gets used
3. the tests were unstable, cache was not always rebuild after wait return null.  So I changed them to force a rebuild.

Please check my commit.

com.unity.cinemachine/Tests/CmCameraProceduralBehaviourCacheTests.cs
```diff
@@ -82,7 +86,8 @@ public IEnumerator AddAndDestroyAllOneByOne()
                 
                 var component = m_CmCamera.gameObject.GetComponent(cmComponent);
                 RuntimeUtility.DestroyObject(component);
-                Assert.True(m_CmCamera.PipelineCacheInvalidated); // invalid pipeline after add
+                Assert.True(m_CmCamera.PipelineCacheInvalidated); // invalid pipeline after destroy
+                yield return null;
```
gaborkb, line 90: Need to wait a frame here, otherwise the Destroyed component might not be destroyed yet.

glabute (approved):

PR #431, opened 2022-04-15
Add CmCamera editor and LensSettings property drawer (still lots of I…

glabute:
Added CmCamera editor (mostly) in UI Elements.
Added LensSettings property drawer (IMGUI).

Any items that need to be drawn in a nonstandard way (e.g. multiple things on a line) are still IMGUI because I can't figure out how to do it in UI Elements.  I am looking for a solution.  We can replace these things gradually.


gaborkb (approved):
There are a couple of things to sort out (alignment, magic padding), but we can merge it and fix later.

PR #432, opened 2022-04-19
Updated asmdef with new 2D URP assembly

unity-cchu:
### Purpose of this PR
This PR added an assembly reference to prevent compile errors with PixelPerfectCamera when using a future version of URP.

### Testing status
Tested against:
* UniversalGraphicsTest_2D
* UniversalGraphicsTest_Foundation

### Technical risk
The change should pose little technical risk or halo effect as only a single reference was added to an asmdef

### DIFF
--- a/com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
+++ b/com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
@@ -1,6 +1,5 @@
 {
     "name": "Cinemachine",
-    "rootNamespace": "",
     "references": [
         "Unity.Timeline",
         "Unity.Postprocessing.Runtime",
--- a/com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
+++ b/com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
@@ -1,5 +1,6 @@
 {
     "name": "Cinemachine",
+    "rootNamespace": "",
     "references": [
         "Unity.Timeline",
         "Unity.Postprocessing.Runtime",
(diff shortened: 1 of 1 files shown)


com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
```diff
@@ -8,7 +8,8 @@
         "Unity.ugui",
         "Unity.RenderPipelines.Universal.Runtime",
         "Unity.2D.PixelPerfect",
-        "Unity.InputSystem"
+        "Unity.InputSystem",
+        "Unity.RenderPipelines.Universal.2D.Runtime"
```
glabute, line 12: Is this a hard dependency or a soft one (other URP dependencies are soft).  What specific part of the code depends on this?

unity-cchu: So the background on this is that I needed to shift the 2D portion of URP into its own assembly. There were no changes to code that would affect anything about how the pixel perfect camera works. However, because of the shift to the new assembly, now Cinemachine can no longer find the PixelPerfectCamera class without adding this new assembly reference (this is code that exists in CinemachinePixelPerfect.cs). If it was a soft dependency before it still is as I've not changed or added anything. Just moved things around.

glabute: Actually it looks as though you've added a hard dependency to Unity.RenderPipelines.Universal.2D.Runtime.  It should be be a soft dependency, because CM is supposed to work with or without URP

unity-cchu: I tested the changes to Cinemachine without URP in my own project, and there were no problems or errors. Seems that all the tests are passing as well.

PR #433, opened 2022-04-20
CMCL-881: overlays for cm3

gaborkb:
### Purpose of this PR

Overlays fix for old freelook.

New overlays for Orbital Follow. Control and select orbits. Orbit selection works only for 3 ring, but could work for sphere too if we want it. The orbit selection directly affects the Vertical Axis value based on the Range and Center values (so it works for sphere too).

### Testing status

- [ ] Added an automated test
- [x] Manually tested on 2021.3.0f1 and 2022.2.0a9

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
@@ -17,9 +17,8 @@ namespace Cinemachine
     [DisallowMultipleComponent]
     [ExecuteAlways]
     [ExcludeFromPreset]
-    [Obsolete("This is deprecated. Use Create -> Cinemachine -> FreeLook camera, or create a virtual camera with " +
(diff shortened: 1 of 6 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -17,7 +17,9 @@ namespace Cinemachine
     [DisallowMultipleComponent]
     [ExecuteAlways]
     [ExcludeFromPreset]
-    [Obsolete("This is deprecated. Use CmFreelook instead.")]
+    [Obsolete("This is deprecated. Use Create -> Cinemachine -> FreeLook camera, or create a virtual camera with " +
```
glabute, line 20: This is a bit TMI.  "This is deprecated. Use Create -> Cinemachine -> FreeLook camera, or create a virtual camera with CinemachineOrbitalFollow body component" would be enough, I think.

glabute (approved):

PR #442, opened 2022-05-05
Dev/yamato fails

gaborkb:

### Purpose of this PR

[JIRA issue. Desc of feature/change. Links to screenshots, design docs, user docs, etc. Remember reviewers may be outside your team, and not know your feature/area that should be explained more.]

### Testing status

- [ ] Added an automated test
- [ ] Manually tested 

### Documentation status

### Technical risk

### Comments to reviewers

### Package version

- [ ] Updated package version


glabute (changes requested):
I don't think you mean to merge to master

gaborkb:
> Can you explain why you moved stuff out of the interface? I'm not too happy having these things in the global scope.

I agree.
But those stuff inside the interface causes one of the yamato fails. We cannot declare structs and stuff inside an interface. However, I may revert the changes, because this fail only happens in 2019 and 2020. I talked to Guillaume about only running yamato test for 2021, 2022, because cm3 won't support 2021 and before anyway.

glabute:
In the worst case, we can make a static class or namespace to hold all these related things as siblings

PR #451, opened 2022-05-21
Dev/cm3 combine vcam targets

glabute:
### Purpose of this PR

- Combine LookAt and Follow targets so that by default only one target is needed.  Optionally, one can add a separate lookAt target by clicking here:
image

- Rename Stages: Body --> PositionControl, Aim --> RotationControl
- Rename CmCamera fields: m_Lens -> Lens, m_Transitions --> Transitions, m_LookAt/m_Follow --> Target


gaborkb (commented):
It does not work well in 2022.1.1f1 and 2022.2.0a10. 

First click on the gear Icon works. But after first click it is very unresponsive. If I do a right-click, then left-click it usually reacts, so it might be something with UI toolkit. It is strange.

glabute:
This is a known ui toolkit regression.  It works in 2021.3.0f1 and again in 2022.2.0a13

gaborkb (approved):

PR #456, opened 2022-05-26
CMCL-0000: Composer bias rounding error.

gaborkb:
### Purpose of this PR
Modifying Screen X and Y could cause a rounding error in Bias X and Y.

### Testing status

- [ ] Added an automated test
- [x] Manually tested


glabute:
I thought we would fix this in master, not in cm3.  In cm3 we will refactor the API and eliminate the problem that way

glabute:
Also I'm not sure this is the right place for the fix.  FramingTransposer probably also suffers from this.  I was thinking that the rounding could be deeper in the screenGuides thingy that both depend on

PR #467, opened 2022-06-06
CMCL-0000: upgrader fixes and refactor

glabute:
I fixed up some upgrader details:
- FreeLookModifier items are only created if there are relevant differences between the rigs

In doing this I noticed that the overall structure could be improved.  I added some notes in CinemachineUpgrader.cs.
The separation of concerns in the various classes needs to be clarified.

Not tested


com.unity.cinemachine/Editor/Upgrader/CinemachineUpgrader.cs
```diff
@@ -754,6 +771,8 @@ public void UpdateTimelineReference(CinemachineVirtualCameraBase upgraded)
                     {
                         var exposedRef = cmShot.VirtualCamera;
                         var vcam = exposedRef.Resolve(director);
+
+                        // This is problematic - what if there were already some null references in the project?
```
gaborkb, line 775: That does not happen, because that cmShot was not collected in m_CmShotsToUpdate.

gaborkb (approved):

PR #471, opened 2022-06-16
CMCL-1003: Sensor size does not affect source control when not using physical camera

gaborkb:
### Purpose of this PR
https://jira.unity3d.com/browse/CMCL-1003

- When not physical camera, sensor size is 0,0. We achieve with before serialize callback.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Package version

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineVirtualCamera.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineVirtualCamera.cs
@@ -56,7 +56,7 @@ namespace Cinemachine
     [ExcludeFromPreset]
     [AddComponentMenu("Cinemachine/CinemachineVirtualCamera")]
     [HelpURL(Documentation.BaseURL + "manual/CinemachineVirtualCamera.html")]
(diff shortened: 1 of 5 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineVirtualCamera.cs
```diff
@@ -622,5 +622,13 @@ internal override bool RequiresUserInput()
                 return true;
             return m_ComponentPipeline != null && m_ComponentPipeline.Any(c => c != null && c.RequiresUserInput);
         }
+
+        public void OnBeforeSerialize()
+        {
+            if (!m_Lens.IsPhysicalCamera) 
+                m_Lens.SensorSize = Vector2.zero;
```
glabute, line 629: Don't use zero!  Use 1, 1 so that calculations won't blow up if it gets deserialized accidentally

gaborkb:
If we'd make LensSettings a class, I could do this whole serialize fix in there, and not everywhere where it is used.

PR #491, opened 2022-07-09
CMCL-0000: bugfix orbit rig handles respect orientation

glabute:
### Purpose of this PR

- Orbit handles were not drawn with the correct orientation when binding mode was LockToTarget and target had its Up rotated away from world up
- Bugfix: OrbitalFollow gizmo was also not drawn with correct orientation in above situation.
- Simplify some overly-verbose code

### Testing status

- [ ] Added an automated test
- [x] Manually tested


gaborkb (approved):
Yamato tests are failing
- splines (they know about it) - probably we should not run tests on those editor versions
- Timeline is missing - not sure why
- CmPixelPerfect has wrong Stage name - fixed on master: 2bc9b167a17f440b0f95201cdc233f013f5ca6ac

glabute:
I'm sure those failing tests have nothing to do with this PR

PR #505, opened 2022-07-20
CMCL-1067: impulse API cleanup

glabute:
CMCL-1067: impulse API cleanup
Remove m_ from public members


gaborkb (approved):
I force pushed to revert an accidental commit to this branch.

PR #523, opened 2022-08-08
CMCL-1118: sample ui dependency removed

gaborkb:
### Purpose of this PR
https://jira.unity3d.com/browse/CMCL-1118

- Added an Assembly Def file for AimingRig, because it depends on UnityEngine.UI.
- AimReticle script is using #if to remove dependency on UnityEngine.UI.
- Removed freelook from scene.

### Testing status
- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


glabute (approved):
A better change (if possible) would be to refactor to use ui toolkit instead of ugui

gaborkb:
> A better change (if possible) would be to refactor to use ui toolkit instead of ugui

Good point! I'll merge it like this, so I can backport it to 2.6, 2.8, 2.9. 
But in CM3, we should upgrade this to use ui toolkit.

gaborkb:
Yamato fails are due to splines package

PR #426, opened 2022-04-11
CMCL-876: Confiner2D confines to midpoint when camera window is bigger than the axis aligned bounding box of the input confiner

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-876
Confiner2D confines to midpoint when camera window is bigger than the axis aligned bounding box of the input confiner.

This case was not cover, and was reported by a user. + code cleanup

### Testing status

[Explanation of what’s tested, how tested and existing or new automation tests. Can include manual testing by self and/or QA. Specify test plans. Rarely acceptable to have no testing.]

- [ ] Added an automated test
- [x] Passed all automated tests
- [x] Manually tested 

### Documentation status

[Overview of how documentation is affected by this change. If there is no effect on documentation, explain why. Otherwise, state which sections are changed and why.]

- [x] Updated [CHANGELOG](https://keepachangelog.com/en/1.0.0/)
- [ ] Updated README (if applicable)
- [ ] Commented all public classes, properties, and methods
- [ ] Updated user documentation

### Technical risk

[Overall product level assessment of risk of change. Need technical risk & halo effect.]

### Comments to reviewers

[Info per person for what to focus on, or historical info to understand who have previously reviewed and coverage. Help them get context.]

### Package version

[Justification for updating either the patch, minor, or major version according to the [semantic versioning](https://semver.org/spec/v2.0.0.html) rules]

- [ ] Updated package version

PR #530, opened 2022-08-10
CMCL-866 - Confiner2D precision issues

gaborkb:
### Purpose of this PR
https://jira.unity3d.com/browse/CMCL-866 (has repro project)
(This fix will need to be added to 3 and 2.8).

- The [initial fix](https://github.com/Unity-Technologies/com.unity.cinemachine/pull/426) for this issue was not precise enough. So now we allow equality for the special case.  
Change at `line 264` in `com.unity.cinemachine/Runtime/Core/ConfinerOven.cs`.

- Added new m_Quality field. 
I played around with the min step size, and it can be used to improve the quality of the confiner in some of these edge cases. I did not find a big performance impact either. I created a prototype quality control, but maybe we should expose the quality parameter as a scripting API only.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Package version


com.unity.cinemachine/Runtime/Core/ConfinerOven.cs
```diff
@@ -236,16 +236,18 @@ public AspectStretcher(float aspect, float centerX)
 
         const long k_FloatToIntScaler = 100000;
         const float k_IntToFloatScaler = 1.0f / k_FloatToIntScaler;
-        const float k_MinStepSize = 0.005f;
+        
+        float m_MinStepSize = k_MinStepSize;
+        const float k_MinStepSize = 0.005f, k_MaxStepSize = 0.000005f;
```
glabute, line 241: max is smaller than min?

com.unity.cinemachine/Runtime/Core/ConfinerOven.cs
```diff
@@ -236,16 +236,18 @@ public AspectStretcher(float aspect, float centerX)
 
         const long k_FloatToIntScaler = 100000;
         const float k_IntToFloatScaler = 1.0f / k_FloatToIntScaler;
-        const float k_MinStepSize = 0.005f;
+        
+        float m_MinStepSize = k_MinStepSize;
```
glabute, line 240: should this be called m_StepSize or something like that?  It's a bit confusing between m_MinStepSize and k_MinStepSize.

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -316,8 +323,9 @@ public void Invalidate()
                 {
                     return false; // input collider is invalid
                 }
-                
-                m_confinerOven = new ConfinerOven(m_OriginalPath, aspectRatio, maxWindowSize);
+
+                quality = (quality - k_QualityMin) / k_QualityMax; // normalize quality to [0-1]
```
glabute, line 327: Normalization is not guaranteed, because caller can pass anything for quality

glabute (changes requested):
Camera still moves (a little) in the sample scene, even at max quality.
I'm not convinced that this quality business is a good strategy for handling such edge cases.  0.000005f is an _insanely_ small step size.  Maybe we can have a discussion about how to solve this better.

gaborkb:
No longer relevant. See clipper2 in CM3

PR #531, opened 2022-08-11
CMCL-1144: Refactor upgrader to iterate scenes less and some bugfixes

gaborkb:
### Purpose of this PR

https://jira.unity3d.com/browse/CMCL-1144:
- Refactor: Upgrader opens scenes less (from 5 to 3)!
- Fixed: `DollyTracks` upgraded to `SplineDolly` could lose their `path` references.
     1. Prefab assets - fixed (by cleaning up at the end, so that references are not lost)
     2. Prefab instances - fixed (by ordering upgrades to always upgrade referencables (e.g. cmPath) first and adding this reference to the converted copy)
     3. Non-prefabs - fixed (by ordering upgrades to always upgrade referencables (e.g. cmPath))
- Fixed: `2DZoom scene` - why is the new `CmCamera` disabled. It was disabled because the gameobject has a `Zoom script` that requires `CinemachineVirtualCamera` to be present.  Vcam's `OnEnable` function detects duplicates and disabled `CmCamera`. We cannot remove `CinemachineVirtualCamera` from the gameobject because of the `RequireComponent attribute`, so it gets disabled and replaced by an enabled `CmCamera`. 
This is also the case, if the script has `[RequireComponent(typeof(CmCamera))]`.
- Error messages in console:
-> Fixed: Our animation track reference updater was too greedy - it overwrote non cm references too! 
(Warning message about setting animation track curves)

### Testing status

- [ ] Added an automated test
- [x] Manually tested above mentioned fixes

### Documentation status


com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -86,8 +86,18 @@ public GameObject UpgradeComponents(GameObject go)
                      orbital.UpgradeToCm3(go.GetComponent<CinemachineOrbitalFollow>());
                      ConvertInputAxis(go, "Look Orbit X", ref orbital.m_XAxis, ref orbital.m_RecenterToTargetHeading);
                 }
+
                 if (ReplaceComponent<CinemachineTrackedDolly, CinemachineSplineDolly>(go))
-                    go.GetComponent<CinemachineTrackedDolly>().UpgradeToCm3(go.GetComponent<CinemachineSplineDolly>());
```
glabute, line 90: Why was this not covered by the ordering of RootUpgradeComponentTypes?  Paths are supposed to get converted first.  What is the usecase that shows this problem?

I don't like this fix because it secretly upgrades 2 objects when this class is supposed to upgrade one object at a time.

com.unity.cinemachine/Runtime/Behaviours/CinemachineClearShot.cs
```diff
@@ -85,7 +85,7 @@ protected internal override void LegacyUpgrade(int streamedVersion)
             {
                 DefaultTarget = new DefaultTargetSettings 
                 { 
-                    Enabled = m_LegacyLookAt != null || m_LegacyFollow != null,
```
glabute, line 88: Why this change?

com.unity.cinemachine/Runtime/Behaviours/CinemachineStateDrivenCamera.cs
```diff
@@ -131,7 +131,7 @@ protected internal override void LegacyUpgrade(int streamedVersion)
             {
                 DefaultTarget = new DefaultTargetSettings 
                 { 
-                    Enabled = m_LegacyLookAt != null || m_LegacyFollow != null,
```
glabute, line 134: Why this change?

com.unity.cinemachine/Runtime/Deprecated/CinemachineTrackedDolly.cs
```diff
@@ -349,8 +349,6 @@ internal void UpgradeToCm3(CinemachineSplineDolly c)
                 case CinemachinePathBase.PositionUnits.Normalized: c.PositionUnits = UnityEngine.Splines.PathIndexUnit.Normalized; break;
             }
             c.SplineOffset = m_PathOffset;
-            if (m_Path != null)
```
glabute, line 352: Why this change?  Paths should have been upgraded first

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -345,34 +179,72 @@ void UpgradePrefabsAndPrefabInstances()
...
+        void UpgradeAllScenes(Dictionary<int, List<ConversionLink>> conversionLinksPerScene)
+        {
             for (var s = 0; s < m_SceneManager.SceneCount; ++s)
             {
+                // 0. Open scene
                 var scene = OpenScene(s);
                 var timelineManager = new TimelineManager(scene);
-                var conversionLinks = conversionLinksPerScene[m_SceneManager.GetScenePath(s)];
-                var allGameObjectsInScene = GetAllGameObjects();
+                var upgradables = GetUpgradeCandidates(scene.GetRootGameObjects());
```
glabute, line 202: I think here is the answer to my other question.  The correct algorithm might have to be to do this whols thing in 2 passes: 
1. All the leaf objects that may be referenced (e.g. paths)
2. All the things that might refer to other upgraded things (e.g. vcams, freeLooks)

If in fact that is the solution, then we will lose the optimization because we'll have to double the number of scene opens

glabute (changes requested):

com.unity.cinemachine/Runtime/Behaviours/CinemachineClearShot.cs
gaborkb: Fixes: [CMCL-1157](https://jira.unity3d.com/browse/CMCL-1157) - we may need to come up with something else. LegacyUpgrade cannot check for null, because that is only allowed on the main thread. However, OnAfterDeserialize is not guaranteed to run on the main thread. When it does not run on main thread we get an error that stops LegacyUpgrade.
(Error message about EnsureRunningOnMainThread)

com.unity.cinemachine/Runtime/Behaviours/CinemachineStateDrivenCamera.cs
gaborkb: Fixes: [CMCL-1157](https://jira.unity3d.com/browse/CMCL-1157) - we may need to come up with something else. LegacyUpgrade cannot check for null, because that is only allowed on the main thread. However, OnAfterDeserialize is not guaranteed to run on the main thread. When it does not run on main thread we get an error that stops LegacyUpgrade.
(Error message about EnsureRunningOnMainThread)

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
gaborkb: I'll take a look

com.unity.cinemachine/Runtime/Behaviours/CinemachineStateDrivenCamera.cs
glabute: I am confused.  What error was happening, and when was it happening?  The Jira ticket doesn't say

gaborkb: The error happens when LegacyUpgrade is **not** called from the main thread. 

OnAfterDeserialize calls LegacyUpgrade. Unfortunately, OnAfterDeserialize is not guaranteed to be called from the main thread.

glabute: Can you be more specific?  Where is the offending null check?

glabute: nvm I understand now :)

glabute: Let's pull this out into a separate PR.  I think we need a different fix.

gaborkb: https://github.com/Unity-Technologies/com.unity.cinemachine/pull/534

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.Data.cs
```diff
@@ -21,7 +21,15 @@ partial class UpgradeObjectToCm3
...
             typeof(CinemachineVirtualCamera),
         };
-        
+
+        /// <summary>
+        /// Any component that may be referenced by vcams or freelooks 
+        /// </summary>
+        public static readonly List<Type> Referencables = new()
+        {
+            typeof(CinemachinePath)
```
glabute, line 30: What about inemachineSmoothPath?

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -113,204 +113,47 @@ public static void UpgradeProject()
...
-                foreach (var director in directors)
+                var upgradables = 
+                    GetUpgradables(scene.GetRootGameObjects(), UpgradeObjectToCm3.Referencables, false);
+                foreach (var upgradable in upgradables)
                 {
-                    if (renames.ContainsKey(director.name)) // search based on guid name
+                    if (PrefabUtility.IsPartOfAnyPrefab(upgradable)) 
+                        continue;
+                    
+                    foreach (var referencable in UpgradeObjectToCm3.Referencables)
```
glabute, line 137: This is weird.  Why are you looping again?  If upgradable contains multiple upgradable components, UpgradeObjectComponents() will handle it.  No need to upgrade multiple times.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -370,13 +220,51 @@ void UpgradePrefabsAndPrefabInstances()
...
+        static List<GameObject> GetAllGameObjects()
+        {
+            var all = (GameObject[])Resources.FindObjectsOfTypeAll(typeof(GameObject));
+            return all.Where(go => !EditorUtility.IsPersistent(go.transform.root.gameObject)
+                && (go.hideFlags & (HideFlags.NotEditable | HideFlags.HideAndDontSave)) == 0).ToList();
+        }
+
+        static void RestoreConvertedCopyReferences(GameObject original, GameObject converted, List<ConversionLink> conversionLinks)
+        {
+            // TODO: if we'll have more components that can reference upgradables, then we'll need to
```
glabute, line 240: Big red hack-flags here.  And what about dolly cart?

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -370,13 +220,51 @@ void UpgradePrefabsAndPrefabInstances()
...
+            // TODO: create some Data in UpgradeObjectToCM3.Data to make this generic and loopable
```
glabute, line 241: I think if we have some general reflection-based code to handle references in any script, this problem will solve itself

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -400,59 +290,98 @@ void UpgradePrefabsAndPrefabInstances()
                             continue; // is a backup copy
 
                         UpgradeObjectComponents(c.gameObject, timelineManager);
+                    }
+
+                    // clean-up now, because we needed obsolete components to restore references
```
glabute, line 295: I'm confused by this.  When did we restore references?

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using Cinemachine.Utility;
+using NUnit.Framework;
```
glabute, line 7: NUnit?  no unit tests here.  We should not have this dependancy

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -86,8 +87,17 @@ public GameObject UpgradeComponents(GameObject go)
...
                 if (ReplaceComponent<CinemachineTrackedDolly, CinemachineSplineDolly>(go))
-                    go.GetComponent<CinemachineTrackedDolly>().UpgradeToCm3(go.GetComponent<CinemachineSplineDolly>());
+                {
+                    var obsoleteDolly = go.GetComponent<CinemachineTrackedDolly>();
+                    var splineDolly = go.GetComponent<CinemachineSplineDolly>();
+                    obsoleteDolly.UpgradeToCm3(splineDolly);
+                    
+                    // set splineDolly spline reference
+                    if (obsoleteDolly.m_Path != null) 
+                        obsoleteDolly.m_Path.TryGetComponent(out splineDolly.Spline);
```
glabute, line 99: Should do it in obsoleteDolly.UpgradeToCm3(), same way that dollycart does it, above

glabute (changes requested):

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.Data.cs
gaborkb: Yes, it was meant to be the base class -> CinemachinePathBase.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
gaborkb: Yes. Sorry! Good catch! 👍  
The reason why I refactored GetUpgradables was to remove this part. Originally, I had a quick hack here just to filter out anything that's not referable.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
gaborkb: Definitely. I won't remove this for now, because the reflection based referencing will obsolete this anyway. Until that I leave the hack here.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
gaborkb: In UpgradeObjectComponents, in the foreach above this foreach

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
gaborkb: I am using assert in the switch statement to throw an error if we ever have a new path type -> unlikely. I could remove the assert or add a throw new NotImplementedException.

gaborkb: The user may have a special path. So maybe throwing the not implemented exception is good. Or we could just ignore those paths.

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
gaborkb: Yes

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
gaborkb: Tested. If user has a custom CinemachinePathBase, then we add a Spline (so that the upgraded trackedDolly can reference it) on it and display error in console saying that their custom thing was not updated. The gameobject also gets the CinemachineDoNotUpgrade tag.

Is this the desired behaviour? I am not sure if the `CinemachineDoNotUpgrade` tag is necessary.

glabute (approved):

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
glabute: I guess it's good enough for now

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -182,6 +189,8 @@ static int FindRig(string path)
         /// <param name="go">The GameObject being upgraded</param>
         public void DeleteObsoleteComponents(GameObject go)
         {
+            if (go == null) return;
```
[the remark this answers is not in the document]
gaborkb: I think I was testing without deleting any components and I got the error here because go was null. Not something that should happen in a real situtation. It does not hurt having it here, but I can remove it.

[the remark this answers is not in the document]
glabute: It doesn't hurt, but it can hide potential bugs.  It's part of general code cleanliness: better to crash early on an unexpected null than to try to find the source of a covered-up problem

PR #532, opened 2022-08-12
CMCL-1059: Freelook and Collider don't work well together when transitioning between rigs

gaborkb:
### Purpose of this PR
Parent issue: CMCL-1057 - has repro project.

The issue was that we were damping the displacement, but when colliding with a wall, this caused a strong overshot towards or away from the target.
I think, we should damp the delta between the `previous corrected camera position` and the `current corrected position` instead. 

<br>
figure
- The red camera indicates where the camera is **without** a collider. 
- The yellow camera indicates where the camera is **with** a collider. 
<br>

TODO: have not tested with normal vcams!
TODO: I noticed that there is a bump when hitting the collider in this implementation that was not there in the original implementation. This is not good!

### Testing status

- [ ] Added an automated test
- [x] Manually tested with repro project and in freelook collider scene

### Documentation status

### Package version
This fix will need to be ported to 2.6, 2.9, 3.0.

### DIFF
--- a/Runtime/Behaviours/CinemachineCollider.cs
+++ b/Runtime/Behaviours/CinemachineCollider.cs
@@ -328,7 +328,7 @@ namespace Cinemachine
 
                     var displacementSqrMagnitude = displacement.sqrMagnitude;
                     var prevDisplacementSqrMagnitude = extra.previousDisplacement.sqrMagnitude;
-                    if (displacementSqrMagnitude > Epsilon || prevDisplacementSqrMagnitude > Epsilon)
+                    if (displacementSqrMagnitude > Epsilon && prevDisplacementSqrMagnitude > Epsilon)
                     {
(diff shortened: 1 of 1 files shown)


glabute (changes requested):
The reason that we damped the displacement and not the camera position is that it's important that the damping NOT apply to user-generated motion (e.g. user moves the mouse, result should not be damped).

gaborkb:
> The reason that we damped the displacement and not the camera position is that it's important that the damping NOT apply to user-generated motion (e.g. user moves the mouse, result should not be damped).

I think I came up with a solution over the weekend. I felt that this is not good solution in principle.

glabute:
I was thinking that a better solution might involve tracking the camera distance from target and making sure that it does not shrink when it should be growing or grow when it should be shrinking.

glabute:
Replaced by https://github.com/Unity-Technologies/com.unity.cinemachine/pull/535

PR #535, opened 2022-08-16
CMCL-1057: collider damping was not robust

glabute:
### Purpose of this PR

CMCL-1057 (has repro project)
When the FreeLook rigs were extreme the collider damping was not handled correctly, causing the camera to move incorrectly in depth, sometimes overshooting the character.

Solution was to redesign collider damping to use the previous camera position, and to set state.DampingBypass to account for its displacement, eliminating spurious rotation damping in the composer.

The new solution works with all collider strategies.

This fix should be backported to 2.6, 2.8, and 3.0

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineCollider.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineCollider.cs
@@ -281,6 +281,8 @@ namespace Cinemachine
             
                 if (m_AvoidObstacles)
                 {
+                    var initialCamPos = state.CorrectedPosition;
+
                     // Rotate the previous collision correction along with the camera
                     var dampingBypass = Quaternion.Euler(state.PositionDampingBypass);
                     extra.previousDisplacement = dampingBypass * extra.previousDisplacement;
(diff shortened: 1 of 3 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineCollider.cs
```diff
@@ -326,16 +328,32 @@ public override float GetMaxDampTime()
...
-                    extra.previousDisplacement = displacement;
                     state.PositionCorrection += displacement;
+
+                    // Adjust the damping bypass to account for the displacement
+                    if (state.HasLookAt && VirtualCamera.PreviousStateIsValid)
+                    {
+                        var dir0 = extra.previousCameraPosition - state.ReferenceLookAt;
+                        var dir1 = state.CorrectedPosition - state.ReferenceLookAt;
+                        if (dir0.sqrMagnitude > Epsilon && dir1.sqrMagnitude > Epsilon)
+                            state.PositionDampingBypass = UnityVectorExtensions.SafeFromToRotation(
```
gaborkb, line 351: Why is PositionDampingBypass in euler angles? It is always used as a quaternion. Every time it is used and saved there is a quat to euler and euler to quat that could be avoided.

gaborkb (approved):
The damping movement felt strange when I was looking at it in the scene view. I thought something is off, but when I look at it in the game view, it feels/looks great!! 🥇 

Damping is tricky, when I tried to implement it I thought any extra forward movement is bad, but actually yours has that and it feels/looks correct and smooth.

com.unity.cinemachine/Runtime/Behaviours/CinemachineCollider.cs
glabute: It's a good question, but that would be an API change.  Maybe we can fix it for CM3.

PR #558, opened 2022-09-08
CMCL-1178: Refactor

gaborkb:
### Purpose of this PR
Refactor upgrader to use the discussed algorithm with a couple of small adjustments. Plus a couple of small fixes in the code that was needed after the refactor (e.g. no null check in timeline reference update).

### Testing status
- [ ] Added an automated test
- [x] Manually tested in 2022.2.0a16

Tested:
🟢 Previously failing cases (manager camera prefabs). 
🟢 Sample scenes.
🟢 Timeline in prefab asset and instance with modification (CinemachineTrack, AnimationTrack).
🟢  TrackedDolly and DollyCart loses reference when its a prefab instance (prefab asset, and non-prefab works) -> TODO: fix.

### Comments to reviewers
- Left the refactored code in CinemachineUpgraderManager.refactor.cs for clarity for the review only. I'll move it back to CinemachineUpgraderManager when PR is approved.
- Timeline inspector seems to be broken a little bit in 2022.2.0a16. For example, AnimationTrack's buttons are not displayed. I let them know on slack.


com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
```diff
@@ -0,0 +1,262 @@
...
+                    if (upgradedObjects.Contains(go))
+                        continue; // Ignore if already converted (this can happen in nested prefabs)
+                    upgradedObjects.Add(go);
+#if CINEMACHINE_TIMELINE
+                    var originalVcam = go.GetComponent<CinemachineVirtualCameraBase>();
+                    var timelineReferences = timelineManager.GetTimelineReferences(originalVcam);
+#endif
+                    var convertedCopy = Object.Instantiate(go);
+                    UpgradeObjectComponents(convertedCopy, null);
+                    m_ObjectUpgrader.DeleteObsoleteComponents(convertedCopy);
```
glabute, line 68: Is there a reason to delete these now?  It will destroy any references to the deleted components.  Best to delete at the end of the process, I think.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
```diff
@@ -0,0 +1,262 @@
...
+        /// <summary>
+        /// For each prefab asset that has any component from filter
+        /// - Upgrade the prefab asset, but do not delete obsolete components
+        /// - Fix timeline references
+        /// - Fix object references
+        /// - Fix animation references
+        /// </summary>
+        /// <param name="upgradeReferencables">
+        /// True, then only referencable prefab instances are upgraded.
+        /// False, then only non-referencable prefab instances are upgraded.</param>
```
glabute, line 103: What if a prefab has both a referencable and a non-referencable?  e.g. dolly vcam + path in the same prefab.  I think you have to revisit the logic in this function to handle those cases correctly.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
```diff
@@ -0,0 +1,262 @@
...
+        void UpgradePrefabAssets(bool upgradeReferencables)
+        {
+            for (var p = 0; p < m_PrefabManager.PrefabCount; ++p)
+            {
+                m_CurrentSceneOrPrefab = m_PrefabManager.GetPrefabAssetPath(p);
+                using var editingScope = new PrefabUtility.EditPrefabContentsScope(m_CurrentSceneOrPrefab);
+                var prefabContents = editingScope.prefabContentsRoot;
+                var hasReferencable = UpgradeObjectToCm3.HasReferencableComponent(prefabContents);
+                if ((!upgradeReferencables || !hasReferencable) && (upgradeReferencables || hasReferencable))
```
glabute, line 112: This conditional made my head explode.  After talking to my therapist it looks like you're just expressing an xor

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -172,217 +153,36 @@ struct UniqueExposedReference
...
         /// Then, restores timeline references to any of these upgraded instances.
         /// </summary>
         /// <param name="conversionLinks">Conversion links for the current scene</param>
         /// <param name="timelineManager">Timeline manager for the current scene</param>
-        void UpgradePrefabInstances(
-            HashSet<GameObject> upgradedObjects, 
-            List<ConversionLink> conversionLinks, TimelineManager timelineManager)
+        /// <param name="upgradeReferencables">
+        /// True, then only referencable prefab instances are upgraded.
+        /// False, then only non-referencable prefab instances are upgraded.</param>
```
glabute, line 172: What if a prefab has both a referencable and a non-referencable? e.g. dolly vcam + path in the same prefab. I think you have to revisit the logic in this function to handle those cases correctly.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -172,217 +153,36 @@ struct UniqueExposedReference
...
+        void UpgradePrefabInstances(HashSet<GameObject> upgradedObjects, List<ConversionLink> conversionLinks,
+            TimelineManager timelineManager, bool upgradeReferencables)
         {
             var allGameObjectsInScene = GetAllGameObjects();
 
             foreach (var conversionLink in conversionLinks)
             {
                 var prefabInstance = Find(conversionLink.originalGUIDName, allGameObjectsInScene);
+                var hasReferencable = UpgradeObjectToCm3.HasReferencableComponent(prefabInstance);
+                if ((!upgradeReferencables || !hasReferencable) && (upgradeReferencables || hasReferencable))
```
glabute, line 183: oh the pain.  use xor

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -392,7 +192,8 @@ void UpgradeAllScenes(Dictionary<int, List<ConversionLink>> conversionLinksPerSc
                 // GML todo: do we need to do this recursively for child GameObjects?
```
glabute, line 192: What about this?  Did you test with a prefab that has a deep structure?

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
```diff
@@ -392,7 +192,8 @@ void UpgradeAllScenes(Dictionary<int, List<ConversionLink>> conversionLinksPerSc
...
                 SynchronizeComponents(prefabInstance, convertedCopy, m_ObjectUpgrader.ObsoleteComponentTypesToDelete);
 #if CINEMACHINE_TIMELINE
-                timelineManager.UpdateTimelineReference(prefabInstance.GetComponent<CmCamera>(), conversionLink);
+                if (timelineManager != null)
+                    timelineManager.UpdateTimelineReference(prefabInstance.GetComponent<CmCamera>(), conversionLink);
```
glabute, line 196: I'm unhappy about the specific knowledge of CmCamera here, it breaks the abstraction.  We'll deal with this in a subsequent refactor.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
```diff
@@ -0,0 +1,262 @@
...
+                using var editingScope = new PrefabUtility.EditPrefabContentsScope(m_CurrentSceneOrPrefab);
+                var prefabContents = editingScope.prefabContentsRoot;
+                var components = new List<Component>();
+                foreach (var type in m_ObjectUpgrader.RootUpgradeComponentTypes)
+                    components.AddRange(prefabContents.GetComponentsInChildren(type, true).ToList());
+                
+                foreach (var c in components)
+                {
+                    if (c == null || c.gameObject == null)
+                        continue; // was a hidden rig
```
glabute, line 244: this is not needed

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
```diff
@@ -0,0 +1,262 @@
...
+                    m_ObjectUpgrader.DeleteObsoleteComponents(c.gameObject);
+                }
+                    
+                var managers =
+                    prefabContents.GetComponentsInChildren<CinemachineCameraManagerBase>();
+                foreach (var manager in managers)
+                {
+                    manager.InvalidateCameraCache();
+                    var justToUpdateCache = manager.ChildCameras;
+                }
```
glabute, line 257: Is this needed?  Feels hacky.

gaborkb: Not, if we'll add an a code in the inspector to invalidate cache on scene dirty/reload.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
gaborkb: Because when all is upgraded. I do a pass to correct references before obsoletes are deleted.

glabute: hmmm, I was wondering what that was for (cough cough hack cough)

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
gaborkb: When a freelook is upgraded, the components will have null refs (that were pointing to the children).

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
gaborkb: Tested it, no problem.

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
glabute: I don't think so, because you have just gotten them from `prefabContents.GetComponentsInChildren` which will not return null objects, and no upgrade has happened since then

gaborkb: UpgradeObjectComponents called UpgradeFreelook which destroys the child rigs.

glabute: When do you call `UpgradeObjectComponents`?

gaborkb: Just under this `if (c == null || c.gameObject == null)`
```
foreach (var c in components)
{
    if (c == null || c.gameObject == null)
        continue; // was a hidden rig
    if (c.GetComponentInParent<CinemachineDoNotUpgrade>(true) != null)
        continue; // is a backup copy

    // Upgrade prefab and fix timeline references
    UpgradeObjectComponents(c.gameObject, timelineManager);
}
```

glabute: Are we looking at the same thing?  This is what I see:
```                foreach (var c in components)
                {
                    if (c == null || c.gameObject == null)
                        continue; // was a hidden rig
                    if (c.GetComponentInParent<CinemachineDoNotUpgrade>(true) != null)
                        continue; // is a backup copy

                    m_ObjectUpgrader.DeleteObsoleteComponents(c.gameObject);
                }```

Comment (gaborkb) [reply com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs]: Ohhh, no we were not looking at the same thing! 😵‍💫 Sorry! You are right, we don't need this check here! Removed!

File: com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs (hunk 9 above)
Comment (gaborkb) [reply com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs:259]: I'll remove this and add a proper fix in the Editor in a separate PR.

File: com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs [hunk 10]
```diff
@@ -0,0 +1,264 @@
...
+                
+                UpgradePrefabInstances(upgradedObjects, conversionLinksPerScene[s], timelineManager, false);
+                
+                var rootObjects = scene.GetRootGameObjects();
+                var upgradables = GetUpgradables(rootObjects, m_ObjectUpgrader.RootUpgradeComponentTypes, true);
+                UpgradeNonPrefabs(upgradables, upgradedObjects, timelineManager);
+                
+                // restore dolly references in prefab instances
+                foreach (var go in upgradables) 
+                    UpgradeObjectComponents(go, null);
```
glabute, line 199: I don't understand why this is here.  Didn't you just do this in UpgradeNonPrefabs()?

glabute (changes requested):
You still need to merge the files

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
glabute: oh I see... this is re-upgrading the prefab instances that were skipped by UpgradeNonPrefabs.  Didn't we upgrade those earlier?  What case is this handling?

gaborkb: This restores any un-restored references to CmPath/Splines in prefab instances. Because we kept obsolete components up to this point, we can use those to correctly link SplineDolly's and SplineCarts to their splines.

This step could be avoided by only upgrading referencables (prefab asset, prefab instance, non prefab), and then non-referencables (prefab asset, prefab instance, non prefab). Right now, it is not exactly like that, so this extra pass at the end is needed.

glabute: So, what specific case is handled by this?  Is it a prefab instances with an override that points to a non-prefab path?

glabute (approved):
Approving for now, but there is still more work to do

com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.refactor.cs
gaborkb: Yes, that for example.

glabute: Here is what I don't understand.  When the prefab instance got upgraded, the spline will have been set to null because the non-prefab path had not been updated yet, so the reference to the old path would have been lost.  Why wasn't it?

glabute: got it, thanks

PR #566, opened 2022-09-16
CMCL-903: New cinemachine icons

gaborkb:
### Purpose of this PR
[CMCL-903](https://jira.unity3d.com/browse/CMCL-903)
Add new Cinemachine Icons everywhere.

Previously, Cinemachine icons were set through the editor. Select script, select script icon, replace with cm logo. This would have been tedious. So I wrote an editor base class that sets the icon for our editors.

### Testing status

- [x] Updated an automated test
- [x] Passed all automated tests - except InputAxisTests, they fail, but they are not related to this PR.
- [x] Manually tested 

### Documentation status


com.unity.cinemachine/Editor/Editors/CinemachineSplineRollEditor.cs
```diff
@@ -0,0 +1,16 @@
+using Cinemachine;
+using Cinemachine.Editor;
+using Editor.Utility;
+using UnityEditor;
+using UnityEngine;
+
+namespace Editor.Editors
```
glabute, line 7: What is this namespace?  Should be Cinemachine.Editor

com.unity.cinemachine/Editor/Editors/Cinemachine3rdPersonAimEditor.cs
```diff
@@ -0,0 +1,9 @@
+using Cinemachine;
+using Editor.Utility;
+using UnityEditor;
+
+namespace Editor.Editors
```
glabute, line 5: wrong namespace

com.unity.cinemachine/Editor/Editors/CinemachineBrainEditor.cs
```diff
@@ -105,7 +106,8 @@ static void DrawBrainGizmos(CinemachineBrain brain, GizmoType drawType)
         [DrawGizmo(GizmoType.Active | GizmoType.InSelectionHierarchy | GizmoType.Pickable, typeof(CinemachineVirtualCameraBase))]
         public static void DrawVirtualCameraBaseGizmos(CinemachineVirtualCameraBase vcam, GizmoType selectionType)
         {
-            const string kGizmoFileName = "Packages/com.unity.cinemachine/Gizmos/cm_logo.png";
+            string kGizmoFileName = ScriptableObjectUtility.CinemachineRelativeInstallPath + "/Editor/EditorResources/Icons/CmCamera/" +
```
glabute, line 109: Why not use the "Packages/..." format?  It's simpler

com.unity.cinemachine/Editor/Editors/CinemachineBrainEditor.cs
```diff
@@ -105,7 +106,8 @@ static void DrawBrainGizmos(CinemachineBrain brain, GizmoType drawType)
...
+                (EditorGUIUtility.isProSkin ? "Dark/" : "Light/") + "VirtualCamera@256.png";
```
glabute, line 110: Why should the scene-view gizmo depend on skin color?

com.unity.cinemachine/Editor/Editors/CinemachineFollowZoomEditor.cs
```diff
@@ -0,0 +1,10 @@
+using Cinemachine;
+using Editor.Utility;
+using UnityEditor;
+using UnityEngine;
+
+namespace Editor.Editors
```
glabute, line 6: wrong namespace

com.unity.cinemachine/Editor/Editors/CinemachineRecomposerEditor.cs
```diff
@@ -0,0 +1,11 @@
+using Cinemachine;
+using Editor.Utility;
+using UnityEditor;
+using UnityEngine;
+
+namespace Editor.Editors
```
glabute, line 6: wrong namespace

com.unity.cinemachine/Editor/Editors/CinemachineSplineCartEditor.cs
```diff
@@ -0,0 +1,16 @@
+using Cinemachine;
+using Cinemachine.Editor;
+using Editor.Utility;
+using UnityEditor;
+
+namespace Editor.Editors
```
glabute, line 6: wrong namespace

com.unity.cinemachine/Editor/Overlays/CinemachineToolbarOverlay.cs
```diff
@@ -144,11 +144,11 @@ class TrackedObjectOffsetTool : CinemachineTool
     /// By default, CinemachineToolSettingsOverlay.customToolbarItems is null.
     /// </summary>
     [Overlay(typeof(SceneView), "Cinemachine Tool Settings")]
-    [Icon("Packages/com.unity.cinemachine/Gizmos/cm_logo.png")]
+    [Icon("Packages/com.unity.cinemachine//Editor/EditorResources/Icons/CmCamera/Dark/VirtualCamera@256.png")]
```
glabute, line 147: How can I see this icon in the editor?  When is it displayed?
Also: this toolbar appears to be empty.  Why do we have it?

com.unity.cinemachine/Editor/Overlays/SplineRollTool.cs
```diff
@@ -23,12 +23,16 @@ void OnEnable()
...
             {
-                image = EditorGUIUtility.IconContent("d_WheelCollider Icon").image, // TODO: create a proper image
+                image = AssetDatabase.LoadAssetAtPath<Texture2D>(GetIconPath()),
                 text = "Roll Tool",
                 tooltip = "Adjust the roll data points along the spline."
             };
         }
 
+        static string GetIconPath() =>
+            ScriptableObjectUtility.CinemachineRelativeInstallPath + "/Editor/EditorResources/Icons/SplineTrack/" +
```
glabute, line 33: Why not "Packages/..." format?

com.unity.cinemachine/Editor/Utility/EditorWithIcon.cs
```diff
@@ -0,0 +1,31 @@
+using Cinemachine;
+using Cinemachine.Editor;
+using UnityEditor;
+
+namespace Editor.Utility
```
glabute, line 5: I have several issues with this.

- Wrong namespace.  Should be Cinemachine.Editor.
- It's public - anyone using it will get a CmCamera icon by default, whether it's appropriate or not
- It knows about our classes - a pattern to be avoided as it's not sustainable.  

I think you should get rid of this base class.

com.unity.cinemachine/Editor/Windows/CinemachineSettings.cs
```diff
@@ -230,8 +230,9 @@ internal static Texture2D CinemachineLogoTexture
             {
                 if (sCinemachineLogoTexture == null)
                     sCinemachineLogoTexture = AssetDatabase.LoadAssetAtPath<Texture2D>(
-                        ScriptableObjectUtility.CinemachineRelativeInstallPath
-                            + "/Editor/EditorResources/cm_logo_sm.png");
+                        ScriptableObjectUtility.CinemachineRelativeInstallPath +
```
glabute, line 233: Should use "Packages..." form

com.unity.cinemachine/Tests/Editor/ScriptableObjectUtilityTests.cs
```diff
@@ -10,15 +11,19 @@ public class ScriptableObjectUtilityTests
         [Test]
         public void CinemachineInstallPathIsValid()
         {
-            var pathToCmLogo = Path.Combine(ScriptableObjectUtility.CinemachineInstallPath, "Editor/EditorResources/cm_logo_sm.png");
```
glabute, line 13: We should get rid of ScriptableObjectUtility.CinemachineInstallPath, it's no longer useful.  It dates from the days when Cinemachine could be installed in the Assets folder

glabute (changes requested):
I don't like this new base class.
We should get rid of it and set the icons in the script metafiles, like before.
If doing it manually is tedious, maybe you could write a script to do the job, or use find-and-replace with a text editor.

gaborkb:
> I don't like this new base class. We should get rid of it and set the icons in the script metafiles, like before. If doing it manually is tedious, maybe you could write a script to do the job, or use find-and-replace with a text editor.

The other problem is that if we set the metafiles directly, then there is no way to set Dark and Light icons. I'll think of another solution.

glabute:
One idea: you could add an icon helper to InpectorUtility, with an enum for icon types.  Inspectors can call it with enum param, but still it wants code.  It feels wrong to be calling SetIconForObject all the time.  Does it dirty the scene?

A better solution would be to tweak the icon's grey color so that it's compatible with both light and dark schemes.  Then just set the metafile.

gaborkb:
> One idea: you could add an icon helper to InpectorUtility, with an enum for icon types. Inspectors can call it with enum param, but still it wants code. It feels wrong to be calling SetIconForObject all the time. Does it dirty the scene?
> 
> A better solution would be to tweak the icon's grey color so that it's compatible with both light and dark schemes. Then just set the metafile.

That's would be great :)

glabute:
Or, if tweaking the colors is not permitted, what about something like defaulting to the dark theme icons, but on startup, if we have a light skin we load the assets and change the colors in-memory.  Will that result in the modified icons being used?  Is something like that possible?

com.unity.cinemachine/Editor/Overlays/CinemachineToolbarOverlay.cs
gaborkb: This is used with Freelook only to select rigs in the scene view.
In 2022.2, I don't know how to open the overlay inspector, but in other versions you can open it with "space" key.

glabute (approved):
Moved and renamed the icon files for simplicity and consistency with handles

PR #567, opened 2022-09-16
CMCL-1187: Import HDRP autofocus compute shader only on demand

glabute:
### Purpose of this PR

CMCL-1187: Opening a non-HDRP project generated shader compilation errors when the Cinemachine package is imported.  This is because the AutoFocus compute shader is HDRP-specific, and there is no support for conditional compilation of compute shaders.

Solution is to prompt the user to import the shader only upon first use of the feature.  Shader was moved to a "~" folder in the package, so it is not imported by default.  Shader gets imported into the Assets folder.  User may move the shader, but not rename it.

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested


gaborkb (approved):
Why can't the user rename the shader? I renamed it but did not see any problem.

glabute:
Because the code finds it by name.  You won't see a problem until you try to create a new AutoFocus camera - then it will want to re-import the shader.

PR #574, opened 2022-09-22
Dev/2.9 freelook forceposition

gaborkb:
### Purpose of this PR
Replace freelook position search with Steepest Descent.

Binary search does not work, because the angles along the spline are not sorted. Instead, the angle function is parabolic, so finding the minimum of the parabola gives us the interpolation **t** we are looking for.

The results are more precise.

I measured performance:
Old method was 0.6 ms.
New (Steepest Descent) is 0.5-0.8 ms.

In my experience the max iteration count could be set to 5, but I wanted to be generous with 10. In my experience every point was found within 5 steps with good precision. The only time we go to 10, is when the steepest descent jumps around the solutions (e.g. x values: 0.3099999 -> 0.31.00001 -> 0.3099999 -> 0.31.00001 ...). So probably we could set it lower.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Package version

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
@@ -451,8 +451,7 @@ namespace Cinemachine
             const int maxIteration = 10;
             const float epsilon = 0.00005f;
             var x = (min + max) / 2f; // initial guess - mid point
-            int i;
-            for (i = 0; i < maxIteration; ++i)
+            for (var i = 0; i < maxIteration; ++i)
             {
                 var angle = AngleFunction(x);
                 var slope = SlopeOfAngleFunction(x);
@@ -472,10 +471,8 @@ namespace Cinemachine
             // approximating derivative using symmetric difference quotient (finite diff)
             float SlopeOfAngleFunction(float input)
             {
-                var fxBehind = GetLocalPositionForCameraFromInput(input - epsilon);
-                var angleBehind = Vector3.SignedAngle(desiredDirection, fxBehind, Vector3.right);
-                var fxAfter = GetLocalPositionForCameraFromInput(input + epsilon);
-                var angleAfter = Vector3.SignedAngle(desiredDirection, fxAfter, Vector3.right);
+                var angleBehind = AngleFunction(input - epsilon);
+                var angleAfter = AngleFunction(input + epsilon);
                 return (angleAfter - angleBehind) / (2f * epsilon);
             }
         }
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
@@ -445,24 +445,21 @@ namespace Cinemachine
             return m_YAxis.Value; // stay conservative
         }
 
-        const float k_Epsilon = 0.00005f;
-        HashSet<float> m_Xs = new HashSet<float> {10,20,30,40,50}; // alloc 5 places, in my tests that was max - might not need, we could also add an iteration count max instead of this check
+        
         float SteepestDescent(float min, float max, Vector3 desiredDirection)
         {
-            m_Xs.Clear();
-            var x = (min + max) / 2f; // midpoint as guess
-            float angle, slope;
-            while (Mathf.Abs(AngleFunction(x)) >= k_Epsilon)
+            const int maxIteration = 10;
+            const float epsilon = 0.00005f;
+            var x = (min + max) / 2f; // initial guess - mid point
+            int i;
+            for (i = 0; i < maxIteration; ++i)
             {
-                angle = AngleFunction(x);
-                slope = SlopeOfAngleFunction(x);
-                if (slope == 0 || m_Xs.Contains(x))
+                var angle = AngleFunction(x);
+                var slope = SlopeOfAngleFunction(x);
+                if (slope == 0 || Mathf.Abs(angle) < epsilon)
                     break; // found best
-
-                m_Xs.Add(x);
                 x = Mathf.Clamp(x - (angle / slope), min, max); // clamping so we don't overshoot
             }
-            Debug.Log(m_Xs.Count);
             return x;
 
             // localFunctions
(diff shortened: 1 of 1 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -439,40 +439,41 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
-                            i1 = i;
-                        }
-                    }
-                }
-                if (Mathf.Sign(a0) == Mathf.Sign(a1))
-                    return i0 * step;
-                float t = Mathf.Abs(a0) / (Mathf.Abs(a0) + Mathf.Abs(a1));
-                return Mathf.Lerp(i0 * step, i1 * step, t);
+                // We need to find the minimum of the angle of function using steepest descent
+                return SteepestDescent(0f, 1f, dir);
```
glabute, line 443: Very nice algorithm.  However, it fails for this type of orbit configuration:

image

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -439,40 +439,41 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
+                    break; // found best
+                x = Mathf.Clamp(x - (angle / slope), min, max); // clamping is needed so we don't overshoot
+            }
+            return x;
+
+            // localFunctions
+            float AngleFunction(float input)
+            {
+                var point = GetLocalPositionForCameraFromInput(input);
+                return Vector3.SignedAngle(desiredDirection, point, Vector3.right);
```
glabute, line 467: Vector3.SignedAngle is noisy when the angle is small - that's why you're getting imperfect results.  Try changing it to UnityVectorExtensions.SignedAngle (also above, in the calling function)

glabute (changes requested):

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
glabute: It's possible that there's no simple solution that solves all cases.  The old algo also failed in this case, so your fix is an improvement.

glabute: Maybe an easy fix is to start with a better first guess, perhaps based on the spline waypoint that's closest to the camera

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -434,44 +434,83 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
+                return (angleAfter - angleBehind) / (2f * epsilon);
+            }
+            // initial guess based on closest line (approximating spline) to point 
+            float InitialGuess()
+            {
+                UpdateCachedSpline();
+                var pb = m_CachedKnots[1]; // point at the bottom of spline
+                var pm = m_CachedKnots[2]; // point in the middle of spline
+                var pt = m_CachedKnots[3]; // point at the top of spline
+                var dbm = SqrDistanceBetweenLineSegmentAndPoint(pb, pm, cameraPos, out var tbm);
```
glabute, line 483: Why don't you use UnityVectorExtensions.ClosestPointOnSegment() ?

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -434,44 +434,83 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
+                var dmt = SqrDistanceBetweenLineSegmentAndPoint(pm, pt, cameraPos, out var tmt);
+
+                var mid = (min + max) / 2f;
```
glabute, line 486: style issue: a little weird to be using min and max here.  Why are you using them?

glabute (changes requested):

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
gaborkb: I could not find it :) I knew there is a function somewhere that does that.

glabute: Your first instinct should be to ask, rather than re-invent the wheel :)

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -434,46 +434,67 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
+                return SteepestDescent(0f, 1f, dir, cameraPos);
             }
             return m_YAxis.Value; // stay conservative
         }
 
+        float SteepestDescent(float min, float max, Vector3 desiredDirection, Vector3 cameraPos)
+        {
+            var rigToCameraMatrix = Matrix4x4.TRS(
+                Follow.position, 
+                GetRig(1).GetCinemachineComponent<CinemachineOrbitalTransposer>().GetReferenceOrientation(State.ReferenceUp),
```
gaborkb, line 452: Is there a better way to get the orientation?

glabute: This way is fine

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
```diff
@@ -434,46 +434,67 @@ float GetYAxisClosestValue(Vector3 cameraPos, Vector3 up)
...
+        float SteepestDescent(Vector3 desiredDirection, Vector3 cameraPos)
+        {
+            var rigToCameraMatrix = Matrix4x4.TRS(
+                Follow.position, 
+                GetRig(1).GetCinemachineComponent<CinemachineOrbitalTransposer>().GetReferenceOrientation(State.ReferenceUp), 
+                Vector3.one);
+            
+            const int maxIteration = 10;
+            const float epsilon = 0.00005f;
+            var x = InitialGuess(rigToCameraMatrix.inverse.MultiplyPoint(cameraPos));
```
glabute, line 457: Consider using Matrix4x4.Inverse3DAffine.  It's faster and always works for a TRS matrix.

glabute (approved):

glabute (approved):

glabute:
The test project might inspire a nice little unit test.

com.unity.cinemachine/Runtime/Behaviours/CinemachineFreeLook.cs
glabute: Wait a minute - I may have been too hasty to approve this.  It's not correct, I think. The objective should be to cancel out the effect of the X axis value, and that's not what you're doing here.

glabute (approved):

PR #585, opened 2022-09-29
CMCL-0000: sphere visualization orbital follow (experiment)

gaborkb:
### Purpose of this PR
Experiment on visualizing sphere for orbital follow (note there is a barely visible transparent sphere):
Screen Shot 2022-09-29 at 4 05 29 PM

Note: we already a visualization with SceneTools to adjust radius.
That looks like this:
Screen Shot 2022-09-29 at 4 07 54 PM

Also, I think we should not draw the visualization when the scene tools are active. We can discuss these with Swap.


glabute:
Lots of problems with this.
- Both wire discs should pass through the camera, representing X and Y axis movement
- They don't pass through the camera if the target screen pos is not centered or if there is a tracked object offset in the composer.
- Handles.DrawWireDisc is nice, it should be used for the 3-ring mode too
- Handles.DrawWireDisc is not affected by Gizmos.Color, need to set Handles.Color
- Big transparent sphere is not helpful

I will push a commit to address these issues

glabute (approved):

PR #586, opened 2022-09-29
CMCL-0000: Dev/freelook forceposition cleanup opt

glabute:
Cleanup and optimization.
Easier to just do it than to write comments.

In tests, step==5 was still taking too long.  Bumped it up to 29.


gaborkb (approved):
Thank you :)

PR #590, opened 2022-10-05
CMCL-0000: rename old samples, begin new ones with Cameron

glabute:
### Purpose of this PR

Old samples were renamed, and new samples with Cameron were started.  Old samples are still there, but labeled as such.  They were reverted to their pre-upgrade 2.X form, with minor script changes to allow them to compile.

image


gaborkb (changes requested):
Some notes:

- In `FreeLook on Spherical Surface` scene, if I just run forward, the character will have a slight rotation when close to the bottom of the sphere. This is caused by rotation damping (`ThirdPersonController.cs`), because there is a point when `qA` and `qB` are not in the same reference system (for example: qA(88.51, 180.00, 180.00), qB(89.87, 0.00, 0.00)).
Both `qA` and `qB` seem to be correct (visually), so slerp introduces the error.
If rotation damping is 0, then this slight rotation disappears. I would set the rotation damping to 0 for this scene.
Also, damping is not applied only when the character is moving, so if I stop at the bottom I'll have a character looking slight to the side forever, if I make little steps, he aligns correctly.
- Camera prefab has a Character Controller script on it, that I think should be removed.

glabute:
Refactored the controller to add strafe support.
Fixed bug in rotation damping (was calling LookRotation incorrectly).
There is still an issue with the sphere scene - the controller is doing something wrong.  Will debug it later..

glabute:
Why do you think the Cameron prefab should not include the CharacterController?

gaborkb:
> Why do you think the Cameron prefab should not include the CharacterController?
I meant this CharacterController that is removed on the prefab instances.
CharacterController

glabute:
I think it's good to have it by default, because it works well in most cases.  It's not good in cases where world up is not constant, like the sphere scene, so I removed it from there.

gaborkb:
When on the bottom of the sphere, the directions get mixed up (left becomes right, right becomes left)

glabute:
Yeah, there is still some issue there.  Will have to debug later.  For now, I'd like to push this PR so we have it for the bug dojo at 10

gaborkb (approved):

PR #592, opened 2022-10-06
CMCL-1211: Add BoxCollider2D support to Confiner2D.

gaborkb:
### Purpose of this PR
- Added BoxCollider2D support for Confiner2D.

### Testing status

- [x] Added an automated test
- [x] Manually tested 

### Documentation status


com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -286,44 +286,61 @@ public void Invalidate()
                 
                 Invalidate();
                 confinerStateChanged = true;
-                
-                Type colliderType = boundingShape2D == null ? null:  boundingShape2D.GetType();
-                if (colliderType == typeof(PolygonCollider2D))
+                switch (boundingShape2D)
```
glabute, line 289: Missing null check.  Get errors if I delete the existing collider

glabute (changes requested):

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -262,6 +262,10 @@ public void Invalidate()
                 float aspectRatio, out bool confinerStateChanged)
             {
                 confinerStateChanged = false;
+                
+                if (boundingShape2D == null)
```
glabute, line 266: Prior to your change, the old code called Invalidate() if boundingShape2D is null.  Why have you changed that?

gaborkb: I reverted it, just to not add a regression accidentally. There may be some cases where invalidate is needed. E.g. user removes boundingShape2D then adds it back but in some modified version. In this case, if we might not re-calculate cache.

PR #595, opened 2022-10-07
CMCL-0000: [TestingDojo] Remove capsule from FreeLook on Spherical Surface scene

gaborkb:
There was a capsule in the scene that does not belong.


glabute:
already fixed locally

PR #596, opened 2022-10-07
CMCL-0000: [TestingDojo] Remove capsule from FreeLook on Spherical Surface scene

gaborkb:
Warning camera is trying to look at itself not always accurate. For example, move cmCamera to its target, and set the Position and Rotation control to null. The warning message is still there without this fix.

### DIFF
--- a/com.unity.cinemachine/Editor/Utility/CmCameraInspectorUtility.cs
+++ b/com.unity.cinemachine/Editor/Utility/CmCameraInspectorUtility.cs
@@ -90,8 +90,10 @@ namespace Cinemachine.Editor
             if (m_NavelGazeMessage != null)
             {
                 CameraState state = Target.State;
-                bool isNavelGazing = Target.PreviousStateIsValid 
-                    && state.HasLookAt() && (state.ReferenceLookAt - state.GetCorrectedPosition()).AlmostZero();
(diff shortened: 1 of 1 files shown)


glabute:
Already fixed it locally

glabute:
I was fooled by the title.  The actual issue this PR addresses was not fixed locally.

com.unity.cinemachine/Editor/Utility/CmCameraInspectorUtility.cs
```diff
@@ -90,8 +90,10 @@ void UpdateCameraStatus()
             if (m_NavelGazeMessage != null)
             {
                 CameraState state = Target.State;
-                bool isNavelGazing = Target.PreviousStateIsValid 
-                    && state.HasLookAt() && (state.ReferenceLookAt - state.GetCorrectedPosition()).AlmostZero();
+                bool isNavelGazing = Target.PreviousStateIsValid && state.HasLookAt() &&
+                    (state.ReferenceLookAt - state.GetCorrectedPosition()).AlmostZero() &&
+                    (Target.GetCinemachineComponent(CinemachineCore.Stage.Body) != null ||
```
glabute, line 95: this condition is too strong.  Should only check for Aim not null.  Body doesn't matter.

PR #598, opened 2022-10-11
CMCL-0000: new Transitions API

glabute:
Transitions API improvements:
- BlendHint is now multi-select flags
- InheritPosition is one of the flags instead of being separate bool
- Added Events foldout to hide the camera activation event

image


gaborkb (approved):
I think, we should provide more info in the Blend Hint Tooltip. 
Everything option does not make sense and might confuse users. It could be noted in the tooltip, that basically the user can mix BlendHint with other strategies, but strategies don't mix (sphere > cylinder > screen space aim). Not sure how to word it though :)

gaborkb (approved):

PR #599, opened 2022-10-11
CMCL-0000: 3rdPersonShooter fixes [CinemachineTestingDojo]

gaborkb:
### Purpose of this PR

- Decoupled AimTarget reticle from 3rdPersonAim. I think 3rdPersonAim should not know about UI (reticle). It only needs to calculate a position where it is aiming at. 
Upgrader should handle this case. We should probably have a deprecated 3rdPersonAim and a new 3rdPersonAim because of this API change.
- 3rdPersonSample scene red dot is not lagging any more and aim is based on AimTarget, which is set by 3rdPersonAim (or any other cmcamera).
- New sample script **SetAimTarget.cs**, which calculates AimTarget's position using an input VcamBase. 
- New sample script **PositionAimTargetReticle.cs**, which calculates the Reticle screen position from AimTarget's position.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


com.unity.cinemachine/Runtime/Behaviours/Cinemachine3rdPersonAim.cs
```diff
@@ -33,14 +33,6 @@ public class Cinemachine3rdPersonAim : CinemachineExtension
         [Tooltip("How far to project the object detection ray")]
         public float AimDistance;
 
-        /// <summary>This 2D object will be positioned in the game view over the raycast hit point, 
-        /// if any, or will remain in the center of the screen if no hit point is 
-        /// detected.  May be null, in which case no on-screen indicator will appear.</summary>
-        [Tooltip("This 2D object will be positioned in the game view over the raycast hit point, if any, "
-            + "or will remain in the center of the screen if no hit point is detected.  "
-            + "May be null, in which case no on-screen indicator will appear")]
-        public RectTransform AimTargetReticle;
```
glabute, line 42: I don't understand why you removed this.  Can you explain?

com.unity.cinemachine/Samples~/Sample Scenes/Shared/Scripts/PositionAimTargetReticle.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    [Tooltip("CmBrain that defines the live camera")]
+    public CinemachineBrain CmBrain;
+    
+    [Tooltip("AimTarget Reticle")]
+    public RectTransform Reticle;
+    void LateUpdate()
+    {
+        if (CmBrain != null && Reticle != null)
+            Reticle.position = CmBrain.OutputCamera.WorldToScreenPoint(transform.position);
```
glabute, line 14: What guarantee is there that this happens after brain has positioned camera?

com.unity.cinemachine/Samples~/Sample Scenes/Shared/Scripts/PositionAimTargetReticle.cs
```diff
@@ -0,0 +1,16 @@
+using Cinemachine;
+using UnityEngine;
+
+public class PositionAimTargetReticle : MonoBehaviour
+{
+    [Tooltip("CmBrain that defines the live camera")]
+    public CinemachineBrain CmBrain;
```
glabute, line 7: Why Brain?  Why not just Camera?

com.unity.cinemachine/Samples~/Sample Scenes/Shared/Scripts/SetAimTarget.cs
```diff
@@ -0,0 +1,28 @@
+using Cinemachine;
+using UnityEngine;
+
+public class SetAimTarget : MonoBehaviour
+{
+    [Tooltip("A transform to hold the position the camera is aiming at in world space")]
+    public Transform AimTarget;
+    
+    [Tooltip("Cinemachine camera that defines the aim target")]
+    public CinemachineVirtualCameraBase CmCamera;
```
glabute, line 10: It's not great to be specifying a vcam here.  It will fail as soon as a second vcam is activated, so is not a strong solution.  Better would be to dynamically obtain the current active vcam and find its 3rdPersonAim, if any

glabute (changes requested):
Think about having a single root GameObject representing the current aim target.  It has one 3D object which will hold the worldSpace position, and it can also have a canvas with a 2D object representing the screenspace position.  One script can pull the current aim target and set the 2D object at the same time.  The whole thing can be a nice prefab, if you can do it without linking to other scene objects (you can).

com.unity.cinemachine/Runtime/Behaviours/Cinemachine3rdPersonAim.cs
glabute: never mind, I get it

glabute (approved):
Much better.  Pushed a small cleanup, made a prefab.

PR #601, opened 2022-10-12
CMCL-1218: implement CM channels instead of using layers

glabute:
### Purpose of this PR

CMCL-1218: Instead of wasting layers for channeling vcams to splitscreen brains, use a dedicated Channel setting.
This has been requested several times on the forums.

image

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low


gaborkb (approved):
I like the channels (do we need this many?).


I don't like the UI.
The name `Custom Output` hiding priority is not clear to me.

I think using the same like with priority would be cleaner
Output Channel [] (using default)
Priority [] (using default)

glabute:
Agree that `Custom Output` is not a great name.
How about `Priority And Channel` ?

gaborkb (approved):
> Agree that `Custom Output` is not a great name. How about `Priority And Channel` ?

It is better.
But I think not having the fold out is better here. And using the previous priority UI for channel too:
Screen Shot 2022-10-12 at 10 51 17 AM

Maybe not, I am not a great UI person. :)

PR #602, opened 2022-10-12
CMCL-0000: display both input s on line if both systems active, also file renames

glabute:
### Purpose of this PR

Bug dojo issue.
If both input systems are active, display both settings when input controller is collapsed (less confusing)

image

Also (unrelated) renamed a couple of property drawers, for name consistency


com.unity.cinemachine/Editor/Editors/InputAxisControllerEditor.cs
```diff
@@ -57,28 +57,43 @@ public override void OnInspectorGUI()
                 }
                 else
                 {
+#pragma warning disable CS0219 // Variable is assigned but its value is never used
```
gaborkb, line 60: There is no need for this #pragma here, because the ifs at line 84 and 89 use the value, so we won't get a warning.
Also, the pragma here with the other #ifs cause a bit of a clutter to my eyes. That's why I checked :)

gaborkb (approved):

PR #603, opened 2022-10-12
CMCL-0000/cleanup attributes, move SaveDuringPlay to its own folder (for easy s…

glabute:
…haring)


gaborkb (changes requested):
These changes break old sample scene: DualTarget, because of the AxisStatePropertyAttribute change.

Error:
Assets/Samples/Cinemachine/3.0.0-pre.2/Old Samples from CM2/Scenes/DualTarget/MoveAimTarget.cs(24,10): error CS0246: The type or namespace name 'AxisStatePropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)

glabute:
good catch

gaborkb (approved):

PR #604, opened 2022-10-12
CMCL-1219: Support "Place Objects At World Origin" preference option

gaborkb:
### Purpose of this PR
[CMCL-1219](https://jira.unity3d.com/browse/CMCL-1219)
Unfortunately, there is no exposed boolean to check if this option is on. The only way to comply with the set preference is to use the ObjectFactory.PlaceGameObject. This will place the object in the exact origin (vector3.zero). So I am checking for that, because there is very little chance that the user's scene view camera is in the exact origin ever.

### Testing status
- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Package version


com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
```diff
@@ -247,6 +250,8 @@ public static LensSettings MatchSceneViewCamera(Transform sceneObject)
                 }
             }
             return lens;
+
+            static bool AtOrigin(Vector3 a) => a.x == 0f && a.y == 0f && a.z == 0f;
```
glabute, line 254: this is a rather complicated way of saying sceneObject.transform.position == Vector3.zero

PR #605, opened 2022-10-12
CMCL-0000: Physical camera fix

gaborkb:
### Purpose of this PR
We could not change to physical camera, because of ResetProjectionMatrix.

Guillaume noted in the bug dojo:
The physical sensor size is set to 1 mm by default. Should we have a 35 mm default instead?

What should we do about it?

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
@@ -1012,14 +1012,14 @@ namespace Cinemachine
                 Camera cam = OutputCamera;
                 if (cam != null)
                 {
-                    cam.ResetProjectionMatrix();
+                    bool isPhysical = cam.usePhysicalProperties;
+                    if (!isPhysical)
+                        cam.ResetProjectionMatrix();
                     cam.nearClipPlane = state.Lens.NearClipPlane;
                     cam.farClipPlane = state.Lens.FarClipPlane;
                     cam.orthographicSize = state.Lens.OrthographicSize;
                     cam.fieldOfView = state.Lens.FieldOfView;
-
-                    bool isPhysical = cam.usePhysicalProperties;
-
+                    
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
```diff
@@ -1012,14 +1012,14 @@ void PushStateToUnityCamera(ref CameraState state)
                 Camera cam = OutputCamera;
                 if (cam != null)
                 {
-                    cam.ResetProjectionMatrix();
+                    bool isPhysical = cam.usePhysicalProperties;
+                    if (!isPhysical)
+                        cam.ResetProjectionMatrix();
```
glabute, line 1017: ResetProjectionMatrix is there for a reason.  You can't just not call it just because camera is physical.

glabute (changes requested):

com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
gaborkb: But it sets the camera to not physical. Do you know why? We are not setting the cam.projectionMatrix anywhere, so I don't understand why it sets the camera back to not physical.

Now I am checking if physical, then restore it after reseting the projection matrix.

glabute: it's because `ResetProjectionMatrix` behaves badly: it resets `usePhysicalProperties` when it shouldn't..  I checked with the camera team.

glabute (approved):

com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
gaborkb: Thanks for checking! :)

PR #610, opened 2022-10-13
CMCL-0000: GroupFraming is different in old and new  [CinemachineTestingDojo]

gaborkb:
### Purpose of this PR
- [x] Initial state was different between old and new, because new did not run when previous frame was invalid.
- [x] Behaviour is still not the same when damping is enabled.  Both look ok. Old one seems to be buggy. We keep new good behaviour. Will fix old later.

Old FramingTransposer has a bug probably. It is not tracking the TargetGroup. It also modifies the Camera's Z value by -1. New one is tracking the TargetGroup position correctly. So I think the new one is correct, the old one has a bug.

What do you think? I send you a project where you can test it easily to see.

### Testing status

- [ ] Added an automated test
- [ ] Manually tested

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
@@ -160,7 +160,7 @@ namespace Cinemachine
                 return;
 
             var group = vcam.FollowTargetAsGroup;
-            if (group == null || group.Sphere.radius < k_MinimumGroupSize)
+            if (group == null)
                 return;
 
             var extra = GetExtraState<VcamExtraState>(vcam);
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
```diff
@@ -160,7 +160,7 @@ public void Reset()
                 return;
 
             var group = vcam.FollowTargetAsGroup;
-            if (group == null || group.Sphere.radius < k_MinimumGroupSize)
+            if (group == null)
```
glabute, line 163: I don't like this fix, because it has a side-effect: it will now attempt to frame a zero-size group, which it should not do.
Maybe a better fix would be to have our own PreviousStateIsValid flag in ExtraData.  Clear it in extra.Reset().

glabute (changes requested):

glabute (approved):

PR #619, opened 2022-10-18
CMCL-1225: Upgrader UX new design, also checks for prefabs

glabute:
### Purpose of this PR

CMCL-1225: provide object and scene upgrade option only if it's safe (i.e. if no prefabs are involved).
Also implemented new UX approximating Swap's design.
Added Upgrade guide (1st draft) and hot-link to it in the upgrader UX.

Still todo: what about animation references?  They should be checked too.

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested


gaborkb (approved):
Pushed a small cleanup commit.

Still todo: what about animation references? They should be checked too.

PR #621, opened 2022-10-19
CMCL-0000: Improve game view dragging behaviour, some cleanup

glabute:
### Purpose of this PR

Game view guides behave better when dragged:
- If dead zone enabled, dead zone bars will only resize dead zone, not move it.
- If dead zone not enabled, dragging the bars will move screen position as expected.

Also some refactoring/cleanup


gaborkb (changes requested):
I can still move the deadzone position if I start from e.g. size x = 0.5 and then pull it to 0. After it reaches 0 the position of the deadzone moves.

glabute:
Can you be mor specific?  I can't make that happen from your description

glabute:
There might be a very slight position movement from precision inaccuracies, but that was always happening, it's not new.

glabute:
If you like, you can create a separate task to address that

PR #630, opened 2022-10-21
CMCL-0000: Cleanup FoldoutWithEnabledButtonPropertyDrawer

glabute:
Remove some hacks, thanks to code from Damian Campeanu.
Also small bugfix: disabled hard guides were hot-highlighting at top/left screen edges.
Another small bugfix: bad indent in InputAxisController inspector if both input systems present


gaborkb (approved):
Thanks for the comments :)

PR #636, opened 2022-10-25
CMCL-0000: SimpleCar Follow sample scene (WIP)

glabute:
Playing around with an asset-less car.
We can use this as a basis for future car samples.

Seems to have revealed a bug in RotationComposer lookahead (lurches sometimes... we'll have to look into it)


gaborkb (approved):
The RotationComposer bugfix fixed the pop. :)

The car is awesome. The controls are fun (turning is a bit sensitive, but we can fine tune that). 🥇 😃 

My first bugfix was for wheel collider. :)

PR #639, opened 2022-10-27
CMCL-0000: Added Brain Update Modes sample scene

glabute:
### Purpose of this PR

Attempt to address the frequent questions about brain update mode.  What is it for?  What does it do?  How do you get rid of the jitter?


gaborkb (approved):
I think this is useful. I like the help message too.

PR #642, opened 2022-11-01
CMCL-1241:  Fix for 2.9.X: Focal Length is not set by CM

gaborkb:
### Purpose of this PR
[CMCL-1240](https://jira.unity3d.com/browse/CMCL-1240): Focal length is not set by Cinemachine. We only set field of view, but that value is overwritten and we need to set focal length instead.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


glabute (commented):
yikes!  Need to fix this on master too

PR #643, opened 2022-11-02
CMCL-1230: Deterministic unit tests [Draft]

gaborkb:
### Purpose of this PR
[CMCL-1230](https://jira.unity3d.com/browse/CMCL-1230)
New TimeInvariant base class for run time tests. 

### Testing status
- [ ] Added an automated test
- [ ] Manually tested 

### Comments to reviewers
[Draft] until tests pass on [Yamato].

Running tests with UTR locally until we find the cause of the instability in HDRP.


com.unity.cinemachine/Tests/Runtime/CmColliderTests.cs
```diff
@@ -88,37 +80,30 @@ public IEnumerator CheckDampingWhenOccluded()
...
             
             Assert.That(originalCamPosition, Is.EqualTo(m_Vcam.State.GetFinalPosition()).Using(Vector3EqualityComparer.Instance));
 
             var obstacle = GameObject.CreatePrimitive(PrimitiveType.Cube);
             obstacle.transform.position = originalCamPosition; // place obstacle so that camera needs to move
 
-            yield return WaitForOnePhysicsFrame();
-            m_Brain.ManualUpdate();
+            yield return WaitForOnePhysicsFrame(); // ensure that moving the collider (obstacle) takes affect
+            yield return UpdateCinemachine();
```
glabute, line 91: I wonder whether these 2 lines can be formalized in a baseclass method instead of always repeating this pattern with the same comment

com.unity.cinemachine/Tests/Runtime/CmColliderTests.cs
```diff
@@ -88,37 +80,30 @@ public IEnumerator CheckDampingWhenOccluded()
...
-                yield return null;
-            } while ((CinemachineCore.CurrentTime - timerStart) < 0.5f);
-            m_Brain.ManualUpdate();
+            yield return WaitForSeconds(0.5f);
+            yield return UpdateCinemachine();
             
             Assert.That(previousPosition, Is.Not.EqualTo(m_Vcam.State.GetFinalPosition()).Using(Vector3EqualityComparer.Instance));
-            Assert.That(new Vector3(0,0,-4.4f), Is.EqualTo(m_Vcam.State.GetFinalPosition()).Using(Vector3EqualityComparer.Instance));
-            
+            Assert.That(new Vector3(0,0,-4.423887f), Is.EqualTo(m_Vcam.State.GetFinalPosition()).Using(Vector3EqualityComparer.Instance));
```
glabute, line 102: I'm really not a fan of these precise numbers.  To check damping, there should be some sort of loop that checks that the thing has moved _towards_ the target every frame, but not all the way until sufficient time has passed.  There need be no precise numbers.

com.unity.cinemachine/Tests/Runtime/CmColliderTests.cs
```diff
@@ -51,35 +49,29 @@ public IEnumerator CheckSmoothingTime()
...
-            var timerStart = CinemachineCore.CurrentTime;
-            yield return WaitForOnePhysicsFrame();
-            do
-            {
-                m_Brain.ManualUpdate();
-                yield return null;
-            } while ((CinemachineCore.CurrentTime - timerStart) < m_Collider.SmoothingTime);
-            
-            m_Brain.ManualUpdate();
             Assert.That(originalCamPosition, Is.EqualTo(m_Vcam.State.GetFinalPosition()).Using(Vector3EqualityComparer.Instance));
```
glabute, line 72: Instead of `Vector3EqualityComparer.Instance` maybe make a member equality comparer instance, with the thresholds explicitly set, and use that.  I think it could be more readable and understandable.  Right now, I don't know what the equality threshold is without hunting around.

glabute (commented):
I like the new base class.  I'll make a little PR with some ideas for improvement.

com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
```diff
@@ -40,97 +37,80 @@ public override void SetUp()
...
             Assert.That(ReferenceEquals(m_Brain.ActiveVirtualCamera, m_Source));
-            yield return null;
             
             // Active target and blend from source to target completely
             m_Target.enabled = true;
-            var timeElapsed = 0f;
-            m_Brain.ManualUpdate();
-            timeElapsed += CinemachineCore.UniformDeltaTimeOverride;
-            while (timeElapsed < k_BlendingTime)
+            var startTime = CinemachineCore.CurrentTimeOverride;
```
glabute, line 51: Since it's the job of `CinemachineRuntimeTimeInvariantFixtureBase` to manage current time, I would add a `CurrentTime` accessor to it, so clients don't need to know the implementation details

glabute: Building on that idea, you can morph the API into something that is frame-based.  Instead of `UpdateCinemachine` you can have `WaitOneFrame`.  You can add variations that control how many frames to wait, and how many physics frames to wait.  Then, clients reason in terms of frames and current time, as managed by `CinemachineRuntimeTimeInvariantFixtureBase`.

PR #649, opened 2022-11-03
CMCL-0000: brain modes sample scene update

glabute:
### Purpose of this PR

Brain modes sample scene now has toggles to enable/disable interpolation on the cars


gaborkb (approved):
I downloaded the samples and WrapAround stopped working. The script was not found by unity somehow. Renaming the script fixed the issue (WrapAround -> WrapAroundA -> WrapAround). Maybe it was just a local issue.

Without the help button this scene is a bit confusing at first. Maybe making the Help button bigger and renaming it to Explanation or something like that could help first time users understand the scene.  Maybe just making the button bigger is enough :)

PR #669, opened 2022-11-21
Improved performance of gizmo curve rendering in 2.9

John-Whigham:
### Purpose of this PR

While profiling slow editor scene view performance in the _Volvo Test Track_ project it was found that a large proportion of the time was being spent rendering the Cinemachine (2.9) curves as part of gizmo rendering.  To improve the performance of such operations some new batch APIs have been added to Unity allowing multiple points to be transformed or multiple lines drawn by a single API call:

**NOTE: This PR must not land before the Unity PR "Added batch versions of Transform.TransformXXX and Gizmo.Drawline APIs [2022.2]" (https://github.cds.internal.unity3d.com/unity/unity/pull/18472) does - it is open purely for review at this time.**

Batch versions of Transform.TransformPoint

- Transform.TransformPoints(Span<Vector3> positions)
- Transform.TransformPoints(ReadOnlySpan<Vector3> positions, Span<Vector3> transformedPositions)- 

Batch version of Transform.InverseTransformPoint

- Transform.InverseTransformPoints(Span<Vector3> positions)
- Transform.InverseTransformPoints(ReadOnlySpan<Vector3> positions, Span<Vector3> transformedPositions)- 

Batch version of Transform.TransformVector

- Transform.TransformVectors(Span<Vector3> vectors)
- Transform.TransformVectors(ReadOnlySpan<Vector3> vectors, Span<Vector3> transformedVectors)- 

Batch version of Transform.InverseTransformVector

- Transform.InverseTransformVectors(Span<Vector3> vectors)
- Transform.InverseTransformVectors(ReadOnlySpan<Vector3> vectors, Span<Vector3> transformedVectors)- 

Batch version of Transform.TransformDirection

- Transform.TransformDirections(Span<Vector3> directions)
- Transform.TransformDirections(ReadOnlySpan<Vector3> directions, Span<Vector3> Directions)- 

Batch version of Transform.InverseTransformDirection

- Transform.InverseTransformDirections(Span<Vector3> directions)
- Transform.InverseTransformDirections(ReadOnlySpan<Vector3> directions, Span<Vector3> Directions)- 

Trunk PR (landed): https://github.cds.internal.unity3d.com/unity/unity/pull/15733
2022.2 PR (under review): https://github.cds.internal.unity3d.com/unity/unity/pull/18472

This PR adds support for these APIs to Cinemachine 2.9 more than doubling the performance of it's gizmo curve rendering.  

A version define CINEMACHINE_USE_BATCH_POINT_LINE_APIS has been added to the editor and runtime asmdefs allowing the new code to only apply to the current 2022.2.0b16 version and later, the original gizmo curve rendering code remains in place for use when the package is used with earlier Unity versions.

### Testing status

Functional tests for the API additions were part of the original PRs.  Manual testing of the Cinemachine changes in this PR was done using one of the sample scenes included with the CM package plus the _Volvo Test Track_ project to make sure that the curves looked the same before and after then change.

image
image
image
image
image
image

A non-automated performance test was used to verify the improvement in scene rendering performance, taking the median of ten redraws in the _Volvo Test Track_ project the scene view render time reduced from 181ms to 89ms.  

Looking more in depth using the Superluminal profiler it was shown that `CinemachineSmoothPathEditor.DrawGizmos()` in this scene reduced from 135ms to 28ms (so over 4.8x the speed)

Backwards compatibility testing of the `CINEMACHINE_USE_BATCH_POINT_LINE_APIS `version define was done by manually changing the asmdef so it only applied to a future version then re-testing the Cinemachine sample scene and Volvo Test Track project to ensure curve rendering was unaffected (apart from performance which was verified to be slower once again)

- [*] Added an automated test
- [*] Passed all automated tests
- [*] Manually tested 

### Documentation status

Is purely a performance improvement, there is no functional change so it wasn't thought that documentation changes were necessary

### Technical risk

Functional tests for the API additions were part of the original PRs.  Manual functional testing of curve rendering was also performed with the Volvo Test Track project as described above to minimise technical risk

There is limited halo risk as only the rendering of curves is affected, not their function or any user interface code

### Package version

- [ ] Updated package version

### DIFF
--- a/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
@@ -1,3 +1,5 @@
+#define MESH_EXPERIMENT
+
 using UnityEditor;
 using UnityEngine;
 using System;
@@ -247,7 +249,7 @@ namespace Cinemachine.Editor
             }
         }
 
-#if CINEMACHINE_USE_BATCH_POINT_LINE_APIS
+#if !MESH_EXPERIMENT && CINEMACHINE_USE_BATCH_POINT_LINE_APIS
         static Vector4[] m_BezierWeights;
         static int m_BezierWeightsResolution = -1;
         static Vector3[] m_StepPoints;
--- a/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
@@ -247,6 +247,7 @@ namespace Cinemachine.Editor
             }
         }
 
+#if CINEMACHINE_USE_BATCH_POINT_LINE_APIS
         static Vector4[] m_BezierWeights;
         static int m_BezierWeightsResolution = -1;
         static Vector3[] m_StepPoints;
@@ -380,5 +381,16 @@ namespace Cinemachine.Editor
 
             Gizmos.color = colorOld;
         }
+#else
+        // !#if CINEMACHINE_USE_BATCH_POINT_LINE_APIS
+        [DrawGizmo(GizmoType.Active | GizmoType.NotInSelectionHierarchy
+             | GizmoType.InSelectionHierarchy | GizmoType.Pickable, typeof(CinemachineSmoothPath))]
+        static void DrawGizmos(CinemachineSmoothPath path, GizmoType selectionType)
+        {
+            var isActive = Selection.activeGameObject == path.gameObject;
+            CinemachinePathEditor.DrawPathGizmo(path,
+                isActive ? path.m_Appearance.pathColor : path.m_Appearance.inactivePathColor, isActive);
+        }
+#endif  // #if CINEMACHINE_USE_BATCH_POINT_LINE_APIS
     }
 }
--- a/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
@@ -1,8 +1,11 @@
 using UnityEditor;
 using UnityEngine;
+using System;
 using System.Linq;
 using System.Collections.Generic;
 using UnityEditorInternal;
+using Cinemachine.Utility;
+using UnityEngine.Assertions;
 
 namespace Cinemachine.Editor
 {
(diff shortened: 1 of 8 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachinePath.cs
```diff
@@ -98,10 +98,17 @@ float GetBoundingIndices(float pos, out int indexA, out int indexB)
             return pos;
         }
 
+#if CINEMACHINE_USE_BATCH_POINT_LINE_APIS
```
glabute, line 101: It's not necessary to have API protectors in the runtime module, this change does not depend on new APIs

com.unity.cinemachine/Runtime/com.unity.cinemachine.asmdef
```diff
@@ -84,7 +84,12 @@
             "name": "com.unity.render-pipelines.high-definition",
             "expression": "14.0.0",
             "define": "CINEMACHINE_HDRP_14"
+        },
+        {
+            "name": "Unity",
+            "expression": "2022.2.0b16",
+            "define": "CINEMACHINE_USE_BATCH_POINT_LINE_APIS"
```
glabute, line 91: Don't need this

com.unity.cinemachine/Editor/Editors/CinemachineSmoothPathEditor.cs
```diff
@@ -244,6 +247,142 @@ void DrawPositionControl(int i, Matrix4x4 localToWorld, Quaternion localRotation
...
+        static ElementType[] EnsureArraySize<ElementType>(ElementType[] array, int requiredSize)
+        {
+            return ((array == null) || (array.Length < requiredSize)) ? new ElementType[requiredSize] : array;
+        }
+
+        [DrawGizmo(GizmoType.Active | GizmoType.NotInSelectionHierarchy
+             | GizmoType.InSelectionHierarchy | GizmoType.Pickable, typeof(CinemachineSmoothPath))]
+        static void DrawGizmos(CinemachineSmoothPath path, GizmoType selectionType)
+        {
+            int numSegments = path.m_Waypoints.Length - (path.Looped ? 0 : 1);
```
glabute, line 267: Why is there a separate implementation for this?  Can't you just call CinemachinePathEditor.DrawPathGizmo?

glabute (commented):
I think you should hold off on this.  I have another idea for optimization that might be even better and doesn't depend on new APIs.  I'm thinking of building and caching a mesh, which can be drawn wireframe in a single call.  The key would be to cache it and only invalidate when necessary.

John-Whigham:
> 

Drawing a mesh may well be faster if it can be cached (I suspect building it each time would not be faster than the code with these new APIs, may or may not be faster than the original method), certainly worth testing. 

The points for the curves using the Gizmo Line APIs could also be cached of course to significantly improve performance with this method (at the expense of memory) but I didn't feel confident enough in my understandings of CM to implement that - particularly knowing when it had to be regenerated

John-Whigham:
Note this PR won't be proceeding, @glabute is going to tackle the performance issue by a different route generating and caching local space meshes for the curve rendering which should give good performance without requiring the new APIs.

glabute:
Implemented by #674

PR #675, opened 2022-11-23
CMCL-1271: Adjust Confiner2D

gaborkb:
### Purpose of this PR
[CMCL-1271](https://jira.unity3d.com/browse/CMCL-1271)
Adjusting confiner to the window size (field of view/ortho size) of the camera is costly, so we don't adjust automatically. But we lacked a simple API for users to adjust it easily. To adjust for just the window size, we don't need a full cache recomputation.
Added AdjustConfiner function and a button to the UI (similar to Recompute Cache).

Fixed gc allocs in IsInsideOriginal.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Comments to reviewers
I should update the confiner2D editor code to use UITK. 

### Package version
- [ ] Updated package version


com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -101,19 +107,26 @@ void Reset()
...
+        /// Adjusts the confiner to the Cinemachine Camera's window size.
+        /// Call this when you change Orthographic Size or Field of View of the Cinemachine Camera's lens.
+        /// Calculating the new confiner is fast, but causes allocations.
+        /// </summary>
+        /// <remarks>
+        /// Most often, it is more efficient to have more Cinemachine Cameras with different lens settings
+        /// that have their own confiners and blend between them instead of changing
+        /// one Cinemachine Camera's lens and calling this over and over.
+        /// </remarks>
+        public void AdjustConfiner() => m_AdjustConfiner = true;
```
glabute, line 120: Can we think of a better name?  "AdjustConfiner" says basically nothing about why you would want to invoke it, and as a user it would just make me feel nervous.

gaborkb: InvalidateComputedConfiner?

I renamed it to that. Cleaned up; there is no need for a flag.

com.unity.cinemachine/Editor/Editors/CinemachineConfiner2DEditor.cs
```diff
@@ -1,22 +1,26 @@
...
-        GUIContent m_ComputeSkeletonLabel = new GUIContent(
+        GUIContent m_ComputeSkeletonLabel = new(
             "Oversize Window", "If enabled, the confiner will compute a skeleton polygon to "
                 + "support cases where camera window size is bigger than some regions of the "
                 + "confining polygon.  Enable only if needed, because it's costly");
         GUIContent m_MaxWindowSizeLabel;
-        GUIContent m_InvalidateCacheLabel = new GUIContent(
-            "Invalidate Cache", "Force a recomputation of the polygon cache.  "
+        GUIContent m_InvalidateComputedConfinerLabel = new(
+            "Invalidate Computed Confiner", "Invalidates the current confiner, so a new one is computed next frame.  " +
```
glabute, line 20: I'm still not happy with the names.  Putting myself in the place of the user, I ask: What is the difference between the computed confiner and the cache?  Cache of what?  The help says "invalidate the current confiner".  What is that?

glabute: It seems to me that there are 2 cached things: the polygon points, and the lens settings.  The names should reflect that, so it's easy to understand which one should be invalidated in what circumstances.

glabute (changes requested):

com.unity.cinemachine/Editor/Editors/CinemachineConfiner2DEditor.cs
```diff
@@ -1,31 +1,39 @@
...
         GUIContent m_MaxWindowSizeLabel;
-        GUIContent m_InvalidateCacheLabel = new GUIContent(
-            "Invalidate Cache", "Force a recomputation of the polygon cache.  "
-                + "This needs to be done if points inside the bounding polygon change");
+        GUIContent m_InvalidateComputedConfinerLabel = new(
+            "Invalidate Computed Confiner", "Invalidates the current confiner, so a new one is computed next frame.\n" +
+            "Call this when when the Field of View or Orthographic Size changes!");
+        GUIContent m_InvalidateFullCacheLabel = new(
+            "Invalidate Bounding Shape Cache", 
+            "Forces a re-computation of the whole confiner2D cache next frame.  This recomputes:\n" +
```
glabute, line 24: This is better, but it's still not doing it for me.  Asks the naive user who has not looked at the implementation: "what is the difference between the current confiner and the whole confiner?"

There should be 2 invalidation commands:
1. InvalidatePolygonCache()
2. InvalidateLensCache()
You call the first one when the polygon changes, you call the second one when the lens changes.
You should use these names, and update the tooltips, docs, method and variable names accordingly.

glabute (changes requested):

com.unity.cinemachine/Editor/Editors/CinemachineConfiner2DEditor.cs
gaborkb: Good names :)

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -101,19 +107,43 @@ void Reset()
             MaxWindowSize = -1;
         }
 
-        /// <summary>Invalidates cache and consequently trigger a rebake at next iteration.</summary>
-        public void InvalidateCache()
+        /// <summary>
+        /// Invalidates the lens cache, so a new one is computed next frame.
+        /// Call this when when the Field of View or Orthographic Size changes!
```
glabute, line 112: ~exclamation point~

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -101,19 +107,43 @@ void Reset()
...
+            var extra = GetExtraState<VcamExtraState>(VirtualCamera);
+            extra.BakedSolution = null;
         }
 
+        /// <summary>
+        /// Invalidates Bounding Shape Cache.
+        /// Forces a costly re-computation of the whole cache.  This recomputes the bounding shape cache, and
+        /// the computed confiner cache.
+        /// Call this when the input bounding shape changes (non-uniform scale, rotation, or
+        /// points are moved, added or deleted)!
```
glabute, line 131: ~exclamation point~

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner2D.cs
```diff
@@ -101,19 +107,43 @@ void Reset()
...
+        /// <remarks>
+        /// It is much more efficient to have more Cinemachine Cameras with different input bounding shapes and
+        /// blend between them instead of changing one Confiner2D's input bounding shape and calling this over and over.
+        /// </remarks>
+        public void InvalidateBoundingShapeCache() => m_ShapeCache.Invalidate();
+        
+        [Obsolete("Call InvalidateBoundingShapeCache() instead.", false)]
+        public void InvalidateCache() => m_ShapeCache.Invalidate();
+
         /// <summary>Validates cache</summary>
```
glabute, line 142: Could we make a better summary?  Does it force recomputation of the cache?

glabute (approved):
Please address the comments before committing.

PR #677, opened 2022-11-24
CMCL-1273: 3D samples

gaborkb:
### Purpose of this PR
[CMCL-1273](https://jira.unity3d.com/browse/CMCL-1273)
Added a couple of 3d sample scenes:

- FadeOutNearbyObjects
- StateDrivenCamera - Roadie Run - [CMCL-921](https://jira.unity3d.com/browse/CMCL-921)
- Clearshot
- Trigger Volume
- Timeline sample (separate because of dependency), we could add more maybe. Not sure.
- ThirdPerson Shooter sample updated. Shooting was not setup. Added a wall.

Dependencies for 2d and 3d samples are not set, because those share dependencies. Shared package dependencies are added in the 2D samples PR - https://github.com/Unity-Technologies/com.unity.cinemachine/pull/665.

[CMCL-583](https://jira.unity3d.com/browse/CMCL-583): Race car sample
- Updated the SimpleCar sample to use a mixing camera to demonstrate a very simplified version of what Adam is talking about in the forum video. The mixing camera mixes between two cmCameras. One setup for a slow car movement (closer to the car with smaller fov), and one setup for a faster car movement (farther away from car with bigger fov).
- Added a stunt jump. When the car jumps through a trigger volume cuts to a static side view camera showing the jump (a lot of car games have such "stunt cameras").

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Comments to reviewers
@AntoineCharton if you have the time, could you please play around with the samples? Let me know what you find :)

### Package version

- [ ] Updated package version


com.unity.cinemachine/Samples~/3D Samples/StateDrivenCamera.unity
```diff
@@ -0,0 +1,823 @@
+%YAML 1.1
```
glabute, line 1: This is the Roadie Run sample, right?

com.unity.cinemachine/Samples~/3D Samples/Trigger Volume.unity
```diff
@@ -0,0 +1,1298 @@
+%YAML 1.1
```
glabute, line 1: I think we can come up with a better concept to show trigger volumes.  This just feels broken.  Character is hard to control, and cameras make no sense.  Original upon which this was based was pretty bad too.

com.unity.cinemachine/Samples~/3D Samples/ClearShot.unity
```diff
@@ -0,0 +1,1608 @@
+%YAML 1.1
```
glabute, line 1: Cameras in this scene start to break when you walk the character around.  Needs tuning or more restrictions on the movement.

com.unity.cinemachine/Samples~/Timeline Sample/ActiveControlsWhenTimelineStops.cs
```diff
@@ -0,0 +1,23 @@
+using System;
```
glabute, line 1: This is very confusing.  It's not clear what is being demoed here.  I think this scene can be replaced by a sample that demonstrates a cutscene.

glabute (changes requested):

gaborkb:
Maybe the timeline cutscene could be merged with the trigger volumes. 

Cameron walks into a room -> trigger volume -> activates timeline cutscene -> cutscene ends -> transition back to player controlled setting.

com.unity.cinemachine/Samples~/3D Samples/ClearShot.unity
```diff
@@ -0,0 +1,1653 @@
...
```
glabute, line 1: image

glabute: I don't like the help object being hidden as a child of the camera.  It should be a top-level object, for clarity.  Same comment applies to all the sample scenes.

glabute: The character should turn around and walk back, or warp back to the starting point when it reaches the end of its track.  Maybe you can use the WrapAround script to make that happen without adding new custom scripts.

com.unity.cinemachine/Samples~/3D Samples/FreeLook Deoccluder.unity
```diff
@@ -212,7 +212,8 @@ Transform:
   m_LocalPosition: {x: 0, y: 1.5035291, z: -4.7685847}
```
glabute, line 212: For this and all the FreeLook scenes: new input controls are not set up.  You need to install both new and legacy inputs and reset the component.

image

It should look like this:
image

com.unity.cinemachine/Samples~/3D Samples/RoadieRun.unity
```diff
@@ -38,21 +38,21 @@ RenderSettings:
   m_ReflectionIntensity: 1
```
glabute, line 38: I think you should add a custom blends asset to this scene, and make the walk->run blend faster than the run->walk blend.

Also consider tuning the camera a little, for the camera angles and because the zolly effect is a little extreme, I think.  We should try to make this look really nice, like something you would actually want in a game.

com.unity.cinemachine/Samples~/3D Samples/ThirdPerson Shooter.unity
```diff
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
```
glabute, line 38: We need to figure out how to make UX that's clickable!

com.unity.cinemachine/Samples~/Timeline Sample/ActiveControlsWhenTimelineStops.cs
glabute: This scene is still horrible.  The help helps but it doesn't help enough to really help.  We need to rethink this scene from the ground up.

glabute (changes requested):

glabute:
> Maybe the timeline cutscene could be merged with the trigger volumes.
What timeline cutscene?  I didn't see it

glabute:
Let's talk separately about the cutscene sample

glabute:
actually, that applies to the 2D samples more than the 2D samples.  I will fix on main, but please check the 3D samples for this problem

com.unity.cinemachine/Samples~/3D Samples/ClearShot.unity
gaborkb: I had a custom script to turn around the player, but did not commit it (it was in my local assets). Pushed it. I did not like the wrap around script it is too abrupt in this case in my opinion.

gaborkb:
> Just noticed another problem. if you import Old Samples from CM2, there are conflicts. You'll need new filenames/metafiles for all scripts that also appear in the old samples

Are we keeping the Old Samples? I thought that's only there for testing and we will remove them.

glabute:
I would like to support keeping them, even if we won't ship them.  Such conflicts make testing very difficult.

glabute (commented):
I like most of the changes you made.  Very nice.
One thing: in the scenes that were converted from the old scenes, the CmCameras have some settings that no longer make sense (e.g. PriorityAndChannel checkbox enabled needlessly, composer hard/dead zones enabled needlessly).  These should be cleaned up before merge.

glabute:
A couple of issues have been revealed by these test scenes:
1. "Priority" field in StateDriven and ClearShot children displayed in the manager's inspector doesn't play well with the PriorityAndChannel checkbox of the children
2. Noise frequency increases with time inthe ThirdPerson scene.  Probable bug somewhere
I will create jira issues for these

glabute (approved):
See previous comments for approval conditions.
Also: roadie-run camera needs serious tuning, but we can do that in a subsequent pass.
Also: help button in 3rdperson scene is unclickable, but we can make a separate issue for that

com.unity.cinemachine/Samples~/3D Samples/SimpleCar/MixCamerasBasedOnSpeed.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using Cinemachine;
+using UnityEngine;
+
+[RequireComponent(typeof(CinemachineMixingCamera))]
+public class MixCamerasBasedOnSpeed : MonoBehaviour
+{
+    public float MaxSpeed;
+    public Rigidbody RigidBodyOfCar;
```
glabute, line 9: So, when I said to you that we should be conscious of making custom sample scripts as generic as possible, this is exactly the kind of thing I was talking about.  Why Car?  Because it's a car in _this_ scene.  You should step back and look at this script from a more usage-agnostic perspective.

PR #683, opened 2022-11-25
CMCL-0000: Move unit test to a package

AntoineCharton:
### Purpose of this PR

Move unit tests to a package to allow package specific testing.

### Technical risk

Low: Infrastructure

### Package version

- [ ] Updated package version


gaborkb (changes requested):
editor Test Failure: PackageTestSuite.PackageValidationTests.PackageHasTests
  A package must have either playmode or editor tests in it. If you are using a different package for your tests, make sure to use the 'relatedPackages' field in your main package's manifest like explained here: https://confluence.hq.unity3d.com/display/PAK/How+to+add+a+test+project+for+your+package
  Expected: greater than 0
  But was:  0

guillaumelevass:
I branched off and trying something. I think we might have to give up on package tests and do only project tests. I don't see any downside. It's more consistent and simplify our yamato file.

PR #687, opened 2022-11-29
CMCL-0000: fix sample player rotation

glabute:
Fix weird rotation in third person sample when moving and rotating past +z axis.
Also: add LockCursor option to character controller, enable it in ThirdPersonShooter.


gaborkb (approved):
There is nothing bound to the Fire command of CmInputAxisController in ThirdPerson Shooter scene.

We could bind Fire1 (that's left mouse click).

PR #690, opened 2022-11-30
CMCL-1281: Add CinemachineConfiner3D, deprecate CinemachineConfiner

glabute:
### Purpose of this PR

CMCL-1281: CinemachineConfier is confusing and complicated.  Deprecate it.
- Add CinemachineConfiner3D, with smooth side-stop damping
- Deprecate CinemachineConfiner
- Upgrader upgrades CinemachineContiner to 2D or 3D versions, depending on settings
- Note: Because we can't determine the target class, we don't upgrade animation tracks that were referencing confiner (probably there are none in the field, anyway)

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status


com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using System.Linq;
 using Cinemachine.Utility;
+using Codice.Client.BaseCommands;
+using Mono.Cecil;
```
gaborkb, line 8: This is not needed. What's this?

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
```diff
@@ -95,19 +97,51 @@ public GameObject UpgradeComponents(GameObject go)
...
+#if CINEMACHINE_PHYSICS || CINEMACHINE_PHYSICS_2D
+            // Hack for Confiner upgrade: 2D or 3D?
+            if (oldType == typeof(CinemachineConfiner))
+                return (b as CinemachineConfiner).UpgradeToCm3_GetTargetType();
+#endif
+            if (ClassUpgradeMap.ContainsKey(oldType))
+                return ClassUpgradeMap[oldType];
+
+            return oldType;
+        }
```
gaborkb, line 143: This may be better:

```
public static Type GetBehaviorReferenceUpgradeType(MonoBehaviour b)
{
#if CINEMACHINE_PHYSICS || CINEMACHINE_PHYSICS_2D
    // Hack for Confiner upgrade: 2D or 3D?
    if (b is CinemachineConfiner c)
        return c.UpgradeToCm3_GetTargetType();
#endif
    var oldType = b.GetType();
    return ClassUpgradeMap.ContainsKey(oldType) ? ClassUpgradeMap[oldType] : oldType;
}
```

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner3D.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        Vector3 ConfinePoint(Vector3 camPos)
+        {
+            var mesh = BoundingVolume as MeshCollider;
+            if (mesh != null && !mesh.convex)
+                return Vector3.zero;
+            return BoundingVolume.ClosestPoint(camPos) - camPos;
+        }
```
gaborkb, line 102: Could also be:

```
Vector3 ConfinePoint(Vector3 camPos)
{
    var mesh = BoundingVolume as MeshCollider;
    return mesh != null && !mesh.convex ? Vector3.zero : BoundingVolume.ClosestPoint(camPos) - camPos;
}
```

gaborkb (changes requested):

com.unity.cinemachine/Editor/Upgrader/UpgradeObjectToCm3.cs
glabute: The new version of VisualStudio sometime adds `using` statements all by itself.  I haven't figured out what causes it, or how to make it stop.  It's very irritating.

glabute: Thanks for catching it

com.unity.cinemachine/Runtime/Behaviours/CinemachineConfiner3D.cs
glabute: Do you think that results in improved performance?
If not, I prefer to leave it as-is, because imo it's more readable that way.

gaborkb: I don't think it results in improved performance.

com.unity.cinemachine/Editor/Editors/CinemachineConfiner3DEditor.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                var t = confiner.BoundingVolume.transform;
+                Gizmos.matrix = Matrix4x4.TRS(t.position, t.rotation, t.lossyScale);
+
+                var colliderType = confiner.BoundingVolume.GetType();
+                if (colliderType == typeof(BoxCollider))
+                {
+                    var c = confiner.BoundingVolume as BoxCollider;
+                    Gizmos.DrawWireCube(c.center, c.size);
+                }
```
gaborkb, line 66: I added this under _line 65_ to visualize when the slow happens.
```
Gizmos.color = Color.red;
Gizmos.DrawWireCube(c.center, c.size - 4f * new Vector3(confiner.SlowingDistance / t.lossyScale.x, 
                                                                            confiner.SlowingDistance / t.lossyScale.y, 
                                                                            confiner.SlowingDistance / t.lossyScale.z));
```

gaborkb (commented):
I like the smooth side stop. But I don't like that I don't have control over the strength of the smoothing. It is always the same.

Also, it is difficult to say when the smoothing stops from the slowing distance parameter. I added a visualization in the Confiner3DEditor for myself, but it is not intuitive I think.

com.unity.cinemachine/Editor/Editors/CinemachineConfiner3DEditor.cs
glabute: The slowing can be controlled by increasing the distance param.  Set it very high, then you can see that the vcam begins to slow much sooner.  We can try to think of a setting that will change the shape of the slowing curve, but it is tricky.  Right now the curve looks like this, so it's hard to see any visible effect except near the end 
image

gaborkb (approved):

PR #697, opened 2022-12-06
CMCL-0000: remove name and metafile collisions.  

glabute:
### Purpose of this PR

Metafiles and script names must not collide with old samples.  Fixed that up in 2D Samples.
Also: other small cleanups in the 2D code and samples


gaborkb (approved):
Very nice help texts! :)

PR #698, opened 2022-12-06
CMCL-1294: Improve sample dependency importer pop-up window

gaborkb:
### Purpose of this PR
[CMCL-1294](https://jira.unity3d.com/browse/CMCL-1294)

Import dependency pop-up is only showed when a dependency is missing.
User has 3 options:
1. Import samples and its dependencies
Samples and **missing** dependencies are imported.
2. Import samples but not its dependencies
Samples are imported without dependencies.

---------------------------

**Removed option to abort samples**
Little hack: Samples cannot be stopped from importing, but we can delete it right away. Samples are not deleted if they were already in the user's project (e.g. they import without dependencies and click on reimport again and select Abort). Shared files are also not deleted if they were already there (e.g. user imported another sample that shared some files with the current import).

---------------------------

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### DIFF
--- a/com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
+++ b/com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
@@ -16,14 +16,9 @@ namespace Cinemachine.Editor
     class SampleDependencyImporter : IPackageManagerExtension
     {
         const string k_CinemachinePackageName = "com.unity.cinemachine";
-        const string k_SampleName = "Samples/Cinemachine/";
-        const string k_DeleteFile = "DeleteSamplesInstructions.txt";
         PackageInfo m_PackageInfo;
         IEnumerable<Sample> m_Samples;
         SampleConfiguration m_SampleConfiguration;
-        AddRequest m_PackageAddRequest;
-        int m_PackageDependencyIndex;
-        List<string> m_PackageDependencies = new ();
         PackageChecker m_PackageChecker = new ();
 
         static SampleDependencyImporter() => PackageManagerExtensions.RegisterExtension(new SampleDependencyImporter());
@@ -66,7 +61,11 @@ namespace Cinemachine.Editor
             configuration = null;
             return false;
         }
-
+        
+        const string k_SampleName = "Samples/Cinemachine/";
+        AddRequest m_PackageAddRequest;
+        int m_PackageDependencyIndex;
+        List<string> m_PackageDependencies = new ();
         void LoadAssetDependencies(string assetPath)
         {
             if (m_SampleConfiguration != null)
@@ -82,11 +81,13 @@ namespace Cinemachine.Editor
                         {
                             // Import common asset dependencies
                             sharedAssetsImported = ImportAssetDependencies(
-                                m_PackageInfo, m_SampleConfiguration.SharedAssetDependencies, out _);
+                                m_PackageInfo, m_SampleConfiguration.SharedAssetDependencies, 
+                                out bool sharedAssetsReimported);
                             
                             // Import sample-specific asset dependencies
                             localAssetsImported = ImportAssetDependencies(
-                                m_PackageInfo, sampleEntry.AssetDependencies, out bool reimport);
+                                m_PackageInfo, sampleEntry.AssetDependencies, 
+                                out bool localAssetsReimported);
                             
                             // Import common amd sample specific package dependencies
                             m_PackageDependencyIndex = 0;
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
```diff
@@ -138,17 +155,71 @@ void ImportPackageDependencies()
...
+            ImportSampleButNotPackageDependencies = 1,
+            AbortImport = 2
+        }
+        static UserChoice PromptUserConfirmation(List<string> dependencies)
+        {
+            return (UserChoice) EditorUtility.DisplayDialogComplex(
+                "Import Sample Package Dependencies",
+                "These samples contain package dependencies that your project does not have: \n" +
+                dependencies.Aggregate("", (current, dependency) => current + (dependency + "\n")),
+                "Import samples and its dependencies",
```
JWalker0007, line 218: I would suggest:

- Import samples and their dependencies
- Import samples without their dependencies
- Cancel importing the samples

com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
```diff
@@ -76,53 +80,66 @@ void LoadAssetDependencies(string assetPath)
...
                             m_PackageDependencies.AddRange(sampleEntry.PackageDependencies);
-                            if (m_PackageDependencies.Count != 0 && PromptUserConfirmation(m_PackageDependencies))
+                            
+                            if (m_PackageDependencies.Count != 0 && 
+                                DoDependenciesNeedToBeImported(out var dependenciesToImport))
                             {
-                                // Import package dependencies using the editor update loop, because
-                                // adding packages need to be done in sequence one after the other
-                                EditorApplication.update += ImportPackageDependencies;
+                                switch (PromptUserConfirmation(dependenciesToImport))
```
glabute, line 100: This code should be before the ImportAssetDependencies part, because if they cancel we don't want to import anything.

com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
```diff
@@ -76,53 +80,66 @@ void LoadAssetDependencies(string assetPath)
...
+                                        break;
+                                    case UserChoice.AbortImport:
+                                        // if samples were already imported, don't delete them
+                                        if (!localAssetsReimported)
+                                            DeleteSampleImported(sample.importPath, 
+                                                !sharedAssetsReimported, m_SampleConfiguration.SharedAssetDependencies,
+                                                localAssetsImported, sampleEntry.AssetDependencies);
+                                        break;
+                                    default:
+                                    case UserChoice.ImportSampleButNotPackageDependencies:
```
glabute, line 117: If the user presses the little x at the top-right of the dialog, what happens?  The default action?  It should be Abort in that case.

glabute (changes requested):

com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
gaborkb: That X is a windows only thing. I don't have that O.O

But according to the docs, X will do the second input option (returning 1) - > ImportSampleButNotPackageDependencies

So I would need to change the order so that AbortImport is 1.

glabute: I'm now of the opinion that we should remove the "Abort" option, since it's kind of a hack and technically it's too late to stop it

com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
gaborkb: I think my hack works reliably so we can leave it in.

I modified the code so that X is abort.

glabute (changes requested):
Cancel doesn't work.  I press Cancel and it imports anyway.  I don't like cancel anymore, because it's not really a cancel.  It will still provoke domain reloads, for importing and then for deleting.  Even if it worked, the whole thing feels wrong.  We should get rid of this option.

gaborkb:
It may be a windows thing then because for me it cancels correctly. I am going to remove the cancel option (and related hack).

gaborkb:
Then that's my fault. Fixable.

Should I remove the Close/abort option?

glabute:
Remove it because as I said, even if it worked it would still feel wrong

PR #708, opened 2022-12-12
CMCL-1306: FindClosestPoint fix

gaborkb:
### Purpose of this PR
[CMCL-1306](https://jira.unity3d.com/browse/CMCL-1306): CinemachinePathBase.FindClosestPoint doesn't search the last segment of its path when there are more than 2 segments.

The problem was that the search radius was clamped to half the radius, but when the path is not looped this may clamp the search radius too short.

For example:
Spline with n points. Not looped.
Search Radius: n. => search whole path even when starting from spline point 0.
However, search radius was clamped to n/2. So now we don't search whole spline just half when starting from 0. 

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### DIFF
--- a/com.unity.cinemachine/Runtime/Deprecated/CinemachinePathBase.cs
+++ b/com.unity.cinemachine/Runtime/Deprecated/CinemachinePathBase.cs
@@ -113,15 +113,14 @@ namespace Cinemachine
             {
                 if (Looped)
                 {
-                    var r = Mathf.FloorToInt(Mathf.Min(searchRadius, (end - start) / 2f));
+                    var r = Mathf.Min(searchRadius, Mathf.FloorToInt((end - start) / 2f));
                     start = startSegment - r;
                     end = startSegment + r + 1;
                 }
                 else
                 {
-                    var r = Mathf.FloorToInt(searchRadius);
-                    start = Mathf.Max(startSegment - r, MinPos);
-                    end = Mathf.Min(startSegment + r + 1, MaxPos);
+                    start = Mathf.Max(startSegment - searchRadius, MinPos);
+                    end = Mathf.Min(startSegment + searchRadius + 1, MaxPos);
                 }
             }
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Runtime/Deprecated/CinemachinePathBase.cs
```diff
@@ -111,13 +111,17 @@ public virtual float StandardizePos(float pos)
...
                 {
-                    start = Mathf.Max(start, MinPos);
-                    end = Mathf.Min(end, MaxPos);
+                    var r = Mathf.FloorToInt(Mathf.Min(searchRadius, (end - start) / 2f));
+                    start = startSegment - r;
+                    end = startSegment + r + 1;
+                }
+                else
+                {
+                    var r = Mathf.FloorToInt(Mathf.Min(searchRadius, end - start));
```
glabute, line 122: what is the purpose of this line?  It seems to do nothing useful, given the next 2 lines

glabute (changes requested):

com.unity.cinemachine/Runtime/Deprecated/CinemachinePathBase.cs
```diff
@@ -111,13 +111,17 @@ public virtual float StandardizePos(float pos)
...
                 {
-                    start = Mathf.Max(start, MinPos);
-                    end = Mathf.Min(end, MaxPos);
+                    var r = Mathf.FloorToInt(Mathf.Min(searchRadius, (end - start) / 2f));
+                    start = startSegment - r;
+                    end = startSegment + r + 1;
+                }
+                else
+                {
+                    var r = Mathf.FloorToInt(searchRadius);
```
glabute, line 122: searchradius is already an int, isn't it?

glabute (changes requested):

glabute (approved):

PR #709, opened 2022-12-14
CMCL-0000: Add cutscene sample

glabute:
### Purpose of this PR

More and better samples.  Renamed some materials.

### Testing status

- [ ] Added an automated test
- [x] Manually tested


glabute:
I'll remove the timeline samples section, I don't think we need it.
And I disagree about disabling the controller during the cutscene.  It adds complexity without adding anything to the scene.

gaborkb (approved):
I would remove the colliders from the loot bunches.

glabute:
yeah, I could clean up a lot of colliders in there

PR #714, opened 2022-12-16
CMCL-0000: cosmetic

glabute:
### Purpose of this PR

~3rdPersonFollow sets LookAt target, just to have something.  Not sure 100% if this is a good idea. 
Also some cosmetic.~
Just cosmetic, no code change

### Testing status

- [ ] Added an automated test
- [x] Manually tested


gaborkb (commented):
I don't think we need to set reference look at in ThirdPersonFollow. 

We removed setting it, because 3rdPersonAim can calculate it.

glabute:
ok, I agree.
Changes now are just cosmetic

glabute:
Merging since changes are just cosmetic

PR #724, opened 2022-12-31
CMCL-1314: Make Yamato green on Cinemachine 2.9

gaborkb:
### Purpose of this PR
[CMCL-1314](https://jira.unity3d.com/browse/CMCL-1314): Make Yamato 🟢 

1. Added missing import call.
2. Replaced deprecated API calls in Unity 2023+. (partially a backport of https://github.com/Unity-Technologies/com.unity.cinemachine/pull/727)
3. Replaced some C# sytax that was unsupported in Unity 2019 and 2020, with a sytax that is supported (new (); -> new List<int>();)
4. Fixed #if guard placement in unit tests - caused guards to not work correctly in yamato remote.
5. Disabled (#if !CINEMACHINE_HDRP) the following unit tests that are unstable on HDRP, because they are not using the TimeInvariant test framework developed on main (CM3): 
     - FreelookForcePositionTests
     - CmColliderTests
     - CamerasBlendingTests

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
+++ b/com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
@@ -1,4 +1,4 @@
-#if false
+#if !CINEMACHINE_HDRP
 using System.Collections;
 using NUnit.Framework;
 using UnityEngine;
--- a/com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
+++ b/com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
@@ -1,3 +1,4 @@
+#if false
 using System.Collections;
 using NUnit.Framework;
 using UnityEngine;
@@ -262,3 +263,4 @@ namespace Tests.Runtime
         }
     }
 }
+#endif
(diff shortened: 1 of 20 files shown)


AntoineCharton:
Looks good. Did you check if we still had to support 2019? I guess we still have to but wanted to make sure.

com.unity.cinemachine/Tests/Runtime/CamerasBlendingTests.cs
```diff
@@ -1,3 +1,4 @@
+#if !CINEMACHINE_HDRP
```
AntoineCharton, line 1: I would suggest using this way instead 
https://github.com/Unity-Technologies/com.unity.cinemachine/blob/release/2.9/Projects/HDRP/Assets/Editor/InitializeTests.cs

https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/reference-attribute-conditionalignore.html

Will still run the test but will be ignored in the pipeline. It will make it easier for us to remember there is a problem here.

com.unity.cinemachine/Tests/Runtime/CmColliderTests.cs
```diff
@@ -1,3 +1,5 @@
+#if !CINEMACHINE_HDRP
+#if CINEMACHINE_PHYSICS
```
AntoineCharton, line 2: See comment on CameraBlendingTest.cs

com.unity.cinemachine/Tests/Runtime/FreelookForcePositionTests.cs
```diff
@@ -1,3 +1,4 @@
+#if !CINEMACHINE_HDRP
```
AntoineCharton, line 1: See comment on CameraBlendingTest.cs

guillaumelevass (approved):

PR #727, opened 2023-01-09
CMCL-0000: Fix deprecated calls in 2023

gaborkb:
### Purpose of this PR
Replaced some deprecated calls with their new version in 2023+.

### Testing status
- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
+++ b/com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
@@ -305,9 +305,13 @@ namespace Cinemachine.Editor
                 return CinemachineCore.Instance.GetActiveBrain(0);
 
             // Create a CinemachineBrain on the main camera
-            Camera cam = Camera.main;
+            var cam = Camera.main;
             if (cam == null)
+#if UNITY_2023_1_OR_NEWER
+                cam = Object.FindAnyObjectByType<Camera>(FindObjectsInactive.Exclude);
+#else
                 cam = Object.FindObjectOfType<Camera>();
+#endif
             if (cam != null)
                 return Undo.AddComponent<CinemachineBrain>(cam.gameObject);
 
--- a/com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
+++ b/com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
@@ -308,7 +308,7 @@ namespace Cinemachine.Editor
             var cam = Camera.main;
             if (cam == null)
 #if UNITY_2023_1_OR_NEWER
-                cam = Object.FindAnyObjectByType<Camera>(FindObjectsInactive.Exclude);
+                cam = Object.FindFirstObjectByType<Camera>(FindObjectsInactive.Exclude);
 #else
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
```diff
@@ -305,9 +305,13 @@ static CinemachineBrain GetOrCreateBrain()
                 return CinemachineCore.Instance.GetActiveBrain(0);
 
             // Create a CinemachineBrain on the main camera
-            Camera cam = Camera.main;
+            var cam = Camera.main;
             if (cam == null)
+#if UNITY_2023_1_OR_NEWER
+                cam = Object.FindAnyObjectByType<Camera>(FindObjectsInactive.Exclude);
```
gaborkb, line 311: The other option would be to call FindFirstObjectOfType. This is slower, but that's what the deprecated call `FindObjectOfType()` did. 

`FindAnyObjectByType` is faster than `FindFirstObjectOfType`, but if more cameras are in the scene, `FindAnyObjectByType` may not find the same on subsequent calls. 

I think we don't need consistency in this case, so I chose the faster method.

AntoineCharton (approved):
Looks good 👍

com.unity.cinemachine/Editor/Menus/CinemachineMenu.cs
glabute: I think the slower version is better in this case.  We do want consistency, or we may end up creating multiple Brains if Camera.main is never set.

glabute (changes requested):
Note that a backport to 2.9 will be necessary.  2.9 must support 2023 also.

com.unity.cinemachine/Runtime/Core/CinemachineVirtualCameraBase.cs
```diff
@@ -801,7 +801,12 @@ class OnDomainReload
         { 
             static OnDomainReload() 
             {
+#if UNITY_2023_1_OR_NEWER
+                var vcams = FindObjectsByType<CinemachineVirtualCameraBase>
+                    (FindObjectsInactive.Include, FindObjectsSortMode.None);
```
glabute, line 806: should use FindFirstObjectOfType here also

com.unity.cinemachine/Runtime/Core/CinemachineVirtualCameraBase.cs
```diff
@@ -801,7 +801,12 @@ class OnDomainReload
         { 
             static OnDomainReload() 
             {
+#if UNITY_2023_1_OR_NEWER
+                var vcams = FindObjectsByType<CinemachineVirtualCameraBase>
+                    (FindObjectsInactive.Include, FindObjectsSortMode.InstanceID);
```
glabute, line 806: Sorry, I was wrong to say earlier that the order needs to be determinate here.  It doesn't.  Better to use the unsorted version.

glabute (changes requested):

glabute (approved):

PR #731, opened 2023-01-16
CMCL-0000: Backport consistent calls from cm3

gaborkb:
### Purpose of this PR
Backport of consistent calls from https://github.com/Unity-Technologies/com.unity.cinemachine/pull/727.

### DIFF
--- a/com.unity.cinemachine/Runtime/Core/CinemachineVirtualCameraBase.cs
+++ b/com.unity.cinemachine/Runtime/Core/CinemachineVirtualCameraBase.cs
@@ -815,7 +815,7 @@ namespace Cinemachine
             {
 #if UNITY_2023_1_OR_NEWER
                 var vcams = FindObjectsByType<CinemachineVirtualCameraBase>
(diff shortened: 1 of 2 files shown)


com.unity.cinemachine/Runtime/Core/CinemachineVirtualCameraBase.cs
```diff
@@ -815,7 +815,7 @@ static OnDomainReload()
             {
 #if UNITY_2023_1_OR_NEWER
                 var vcams = FindObjectsByType<CinemachineVirtualCameraBase>
-                    (FindObjectsInactive.Include, FindObjectsSortMode.None);
+                    (FindObjectsInactive.Include, FindObjectsSortMode.InstanceID);
```
glabute, line 818: Sorry, I was wrong to say earlier that the order need to be determinate here.  It doesn't.  Better to use the faster unsorted version.

glabute (changes requested):

glabute (approved):

PR #743, opened 2023-01-20
CMCL-1315: damping stability for uneven framerates

glabute:
### Purpose of this PR

CMCL-1315: CM's damping is highly sensitive to uneven frame rates, and this can lead to jitter. CINEMACHINE_EXPERIMENTAL_DAMPING is there to address this, and it works, but at an unacceptable performance cost. 
- This PR significantly improves the performance of the stable damping algorithm (replaces a loop with a fixed calculation).  
- Also, the stable algorithm took longer to damp, so its behaviour was noticeably different from the standard algorithm.  Researched a solution for this and added some code to compensate for it, resulting in the same damping time for both algorithms.
- Added StableDampFloat test, to ensure that behaviour is consistent across both methods.

The code is still opt-in, based on the CINEMACHINE_EXPERIMENTAL_DAMPING define.  We might consider changing that.

Smoothing isn't perfect, but it's noticeably better with the stable version - especially when following fast-moving objects.  

Standard damping, with wildly fluctuating framerate:
UnstableDamping

Stable damping, with wildly fluctuating framerate:
StableDamping

For the damp time compensation, I sampled it and used google sheets to fit a curve.
image

### Testing status

- [x] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### DIFF
--- a/com.unity.cinemachine/Runtime/Core/Predictor.cs
+++ b/com.unity.cinemachine/Runtime/Core/Predictor.cs
@@ -139,7 +139,7 @@ namespace Cinemachine.Utility
         /// <returns>The damped amount.  This will be the original amount scaled by
         /// a value between 0 and 1.</returns>
         // Internal for testing
-        public static float StandardDamp(float initial, float dampTime, float deltaTime)
+        internal static float StandardDamp(float initial, float dampTime, float deltaTime)
         {
             if (dampTime < Epsilon || Mathf.Abs(initial) < Epsilon)
                 return initial;
@@ -164,7 +164,7 @@ namespace Cinemachine.Utility
         /// <returns>The damped amount.  This will be the original amount scaled by
         /// a value between 0 and 1.</returns>
         // Internal for testing
-        public static float StableDamp(float initial, float dampTime, float deltaTime)
+        internal static float StableDamp(float initial, float dampTime, float deltaTime)
         {
             if (dampTime < Epsilon || Mathf.Abs(initial) < Epsilon)
                 return initial;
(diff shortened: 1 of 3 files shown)


com.unity.cinemachine/Runtime/Core/Predictor.cs
```diff
@@ -147,5 +129,161 @@ public static Vector3 Damp(Vector3 initial, float dampTime, float deltaTime)
...
+        /// Get a damped version of a quantity.  This is the portion of the
+        /// quantity that will take effect over the given time.
+        /// 
+        /// This is a special implementation that attempts to increase visual stability 
+        /// in the context of an unstable framerate.
+        /// 
+        /// It relies on AverageFrameRateTracker to track the average framerate.
+        /// </summary>
+        /// <param name="initial">The amount that will be damped</param>
+        /// <param name="dampTime">The rate of damping.  This is the time it would
```
gaborkb, line 161: This does not seem to be accurate.
In our UnitTest, standardDamp reaches ~99%, where as stableDamp reaches ~85%

gaborkb (changes requested):
I think a slight doc clarification may be needed, or they may be an error in the algorithm.

com.unity.cinemachine/Runtime/Core/Predictor.cs
glabute: > This does not seem to be accurate. In our UnitTest, standardDamp reaches ~99%, where as stableDamp reaches ~85%

Because you didn't set Damper.AverageFrameRateTracker.SetDampTimeScale

gaborkb: Ok. I understand now. 👍

gaborkb (approved):

PR #751, opened 2023-01-27
CMCL-0000: ThirdPersonShooter update

gaborkb:
### Purpose of this PR
The scene now uses a StateDrivenCamera that switches between ThirdPersonAim (default) and RoadieRun (sprinting) cameras based on Player state.

Also added an impulse listener to the ThirdPersonAim and an Impulse when the gun is fired. The impulse pushes the camera back slightly.

When no 3rdPersonAim is present, PositionAimTargetAndReticle calculates a random target based on the camera forward vector.

### Testing status
- [ ] Added an automated test
- [x] Manually tested


glabute (changes requested):
I think it needs some more tuning.  The blend is weird, especially run->walk.  Something strange is happening - it looks like it goes in and out.  Also, the reticle does strange things, what's up with that?  And the blend happens when running + jumping, which it shouldn't.

com.unity.cinemachine/Samples~/Shared Assets/Scripts/SimplePlayerShoot.cs
```diff
@@ -56,6 +63,7 @@ void Update()
                 var go = Instantiate(BulletPrefab, transform.position + fwd, Quaternion.LookRotation(fwd, transform.up));
                 if (go.TryGetComponent<SimpleBullet>(out var b))
                     b.Fire(fwd, BulletSpeed);
+                ImpulseSource.GenerateImpulseWithForce(ImpulseForce);
```
glabute, line 66: what if ImpulseSource is null?

com.unity.cinemachine/Samples~/Shared Assets/Scripts/PositionAimTargetAndReticle.cs
```diff
@@ -47,6 +47,14 @@ void SetAimTarget(CinemachineBrain brain)
                         if (AimTargetIndicator != null)
                             AimTargetIndicator.position = brain.OutputCamera.WorldToScreenPoint(transform.position);
                     }
+                    else
+                    {
+                        // Emulate unstable aim spread
```
glabute, line 52: Can you explain the purpose of this?

glabute (changes requested):

com.unity.cinemachine/Samples~/Shared Assets/Scripts/PositionAimTargetAndReticle.cs
gaborkb: RoadieRun camera does not have 3rdPersonAim. So when RoadieRun is live, then the aim target is not updated. So aim target remains at the last position 3rdPersonAim set it. This is definitely not good.

When the player is sprinting, their aim is either unstable or blocked. In GearsOfWar's RoadieRun, shooting is disabled when RoadieRunning.

I chose to make shooting spread all over a bit when sprinting instead of disabling shooting. I felt it's less complicated for the user this way.

glabute: It doesn't work properly after you run any significant distance.

glabute: In any case I'm wondering if this is the right place for this logic.  The job of this class is to position the reticle where it's told, but you are abusing it to define game logic.  It's not good architecture.

com.unity.cinemachine/Samples~/Shared Assets/Scripts/SimplePlayerShoot.cs
glabute: I don't like this.  You should create a fire event and set this up in the inspector, not bake Impulse into the code.

com.unity.cinemachine/Samples~/Shared Assets/Scripts/PositionAimTargetAndReticle.cs
gaborkb: What do you mean?

gaborkb: I see now

gaborkb: Now it's fixed.

com.unity.cinemachine/Samples~/Shared Assets/Scripts/SimplePlayerShoot.cs
gaborkb: Fixed

com.unity.cinemachine/Samples~/Shared Assets/Scripts/PositionAimTargetAndReticle.cs
glabute: It still isn't right.  The spread is not correct.  I'm gong to jump on this and fix it, as well as the Fire event thing.

gaborkb: What do you suggest?

Should I not set any value, or not a randomized value?
I could revert this change, and disable shooting while sprinting instead.

glabute: See my PR https://github.com/Unity-Technologies/com.unity.cinemachine/pull/765

glabute (approved):

PR #756, opened 2023-02-01
CMCL-1358: very rare axis drift

glabute:
### Purpose of this PR

CMCL-1358: Very difficult to reproduce, but real.  See detailed description in Jira issue.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### Comments to reviewers

This is difficult to test because the problem is extremely hard to reproduce in the first place.  It's an edge-case condition.  I've given the fix to the original poster on the forum, who is more easily able to repro.  They will confirm the fix.  I think the testing here should just be a manual reading of the code, and a checking that nothing was broken by the change.

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased] 
 - Bugfix: AxisState was not respecting timescale == 0
+- Bugfix: Very occasional axis drift in SimpleFollow when viewing angle is +-90 degrees.
 
 
 ## [2.9.5] - 2023-01-16
--- a/com.unity.cinemachine/Runtime/Components/CinemachineTransposer.cs
+++ b/com.unity.cinemachine/Runtime/Components/CinemachineTransposer.cs
@@ -278,10 +278,10 @@ namespace Cinemachine
                         * targetOrientation).eulerAngles;
                     for (int i = 0; i < 3; ++i)
                     {
+                        if (relative[i] > 180)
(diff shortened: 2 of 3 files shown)


com.unity.cinemachine/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased] 
 - Bugfix: AxisState was not respecting timescale == 0
+- Bugfix: Very occasional axis drift in SimpleFollow when viewing angle is +-90 degrees.
```
gaborkb, line 9: The bugfix message is not correct based on the jira ticket description. You said that small movements are the culprit not the viewing angle.

gaborkb (approved):
Changelog text is outdated I think.

com.unity.cinemachine/CHANGELOG.md
glabute: It is correct.  It has to do with the viewing angle crossing the 90 degree boundary.

gaborkb: From jira ticket:
"I was able to trigger this to happen with small mouse movements from any angle not just -90/90. So it seems like the angle is irrelevant other than the fact that if the bug is triggered the drift happens when viewing from the side and it won’t drift on its own when its near 0/180. Seems like just doing very small mouse movement rotations was enough to trigger it."

But I am ok with the changelog message, if you say it is correct. :)

glabute: yeah, he meant "player angle in worldspace" which he at first thought was relevant, but it isn't

PR #778, opened 2023-02-09
CMCL-0000: Add pooling to tps sample

AntoineCharton:
### Purpose of this PR

Add object pooling to tps sample to avoid unnecessary GC therefore allowing someone to take our code as is without needing optimizations.

Also added some quick optimization to avoid casting on null checks.

Feb-09-2023 14-49-52

Instead of destroying older bullets it disable them and enables them once you fire. Allowing to avoid destroying and instantiating new one every time.

### Testing status

Opened the ThirdPerson Shooter sample and made sure it was working the same. 

### Technical risk

Low sample change.

### Comments to reviewers

I really would like to have this code in the samples. Wether we use it or not in the TPS project or just as a low end device friendly alternative or a commented section in the original code. Instantiating and Destroying can have a big impact on performances.

### DIFF
--- a/com.unity.cinemachine/Samples~/3D Samples.meta
+++ b/com.unity.cinemachine/Samples~/3D Samples.meta
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: b3b631701960b4ffa9d7a96890bdf47f
+guid: 8df9c796287304993be83d53afc4c62d
 folderAsset: yes
 DefaultImporter:
   externalObjects: {}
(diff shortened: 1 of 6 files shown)


glabute:
This is nice, but I'd like to hack at it a bit.  Stay tuned.

glabute:
Did a small refactor to enable the possibility of firing things other than SimpleBullet.  I think it's cleaner this way.  Also disabled player input until help text is dismissed (so it doesn't fire an initial bullet for nothing)

AntoineCharton:
Looks great!

PR #779, opened 2023-02-09
CMCL-0000: Upgrader fixes - gigaya test

gaborkb:
### Purpose of this PR
Found a few missing null checks while testing gigaya.

The other errors are not related to us. They are triggered by us, because gigaya is full of corrupt prefabs. When we open them to check if need to upgrade them, they throw errors. 
The scene that we can't open is also a scene that's corrupted.
These other errors can be cleared in the console.

### Testing status

- [ ] Added an automated test
- [x] Manually tested with Gigaya

### DIFF
--- a/com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
+++ b/com.unity.cinemachine/Editor/Upgrader/CinemachineUpgradeManager.cs
@@ -393,6 +393,8 @@ namespace Cinemachine.Editor
                 
                 foreach (var c in components)
                 {
+                    if (c == null)
+                        continue; // ignore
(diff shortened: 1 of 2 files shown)


glabute (approved):

gaborkb:
> Could we add try/catch to gracefully handle the corrupt things?

I tried that but unfortunately these errors are not caught, because are thrown elsewhere - 
The scene is opened, but the scene itself contains things that throw errors.

For example, this does not catch any errors. 
```
Scene? OpenScene(int sceneIndex)
{
  try
  {
      var path = m_SceneManager.GetScenePath(sceneIndex);
      m_CurrentSceneOrPrefab = path;
      return EditorSceneManager.OpenScene(path, OpenSceneMode.Single);
  }
  catch
  {
      return null;
  }
}
```

PR #781, opened 2023-02-10
CMCL-1342: Change all samples default input system action map to ours

AntoineCharton:
### Purpose of this PR

Use our input action map instead of the default built in one

### Testing status

Opened the samples with and without input system

- [ ] Manually tested 

### Technical risk

Low samples

### Comments to reviewers

Updated some scenes that probably lost serialization at some point.

### DIFF
--- a/com.unity.cinemachine/Samples~/3D Samples.meta
+++ b/com.unity.cinemachine/Samples~/3D Samples.meta
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 8df9c796287304993be83d53afc4c62d
+guid: 5977a87e86308477980ad364b3224827
 folderAsset: yes
 DefaultImporter:
   externalObjects: {}
--- a/com.unity.cinemachine/Samples~/Shared Assets/CinemachineInputActions.inputactions
+++ b/com.unity.cinemachine/Samples~/Shared Assets/CinemachineInputActions.inputactions
@@ -2,7 +2,7 @@
     "name": "CinemachineInputActions",
     "maps": [
         {
-            "name": "Player",
+            "name": "Player1",
             "id": "df70fa95-8a34-4494-b137-73ab6b9c7d37",
             "actions": [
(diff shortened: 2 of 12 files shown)


com.unity.cinemachine/Samples~/3D Samples.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
```
glabute, line 1: There shouldn't be metafiles for the folders in this directory.  If they're here, it's a mistake.  You shouldn't be committing new ones.

com.unity.cinemachine/Samples~/Shared Assets/CinemachineInputActions.inputactions
```diff
@@ -2,7 +2,7 @@
     "name": "CinemachineInputActions",
```
glabute, line 2: We need to think a little more carefully about the design of our input action asset.  We should make standard stuff like move, look, fire, jump, sprint.  We shouldn't have things like "mouse move", "middle click" and "right click".  Also, Player1 and Player2 should have the same actions (though mapped differently).

image

com.unity.cinemachine/Samples~/Shared Assets/Prefabs/CapsuleCar.prefab
```diff
@@ -35,7 +35,7 @@ Transform:
   - {fileID: 4228984098192203911}
```
glabute, line 35: This asset uses a mix of our input actions and the  old ones.  We should check everything and use our stuff only.

glabute (changes requested):

com.unity.cinemachine/Samples~/3D Samples.meta
AntoineCharton: Fixed

glabute:
Implemented in PR #824

PR #801, opened 2023-02-24
CMCL-1399: improve BrainUpdateModes sample - blendlist camera

glabute:
[CMCL-1399](https://jira.unity3d.com/browse/CMCL-1399):
SmartUpdate screen now automatically switches between targets, using a BlendList camera.,


gaborkb (approved):
Very cool to add Blendlist camera here. After your discussion on slack, I was thinking about adding it somewhere, but this is a very good place to add it!

glabute:
I don't know why I didn't think of it before

PR #810, opened 2023-02-27
CMCL-1373: Handling input smarter (2)

gaborkb:
### Purpose of this PR
[CMCL-1373](https://jira.unity3d.com/browse/CMCL-1373): 
Unfortunately, there is no one API for checking what the input type to read is. I could not break the current solution, but it feels fragile nevertheless.

Maybe it is better to only support Vector2.

Slack discussion with input team: https://unity.slack.com/archives/C09Q7LYP9/p1676555716510299
Original PR: https://github.com/Unity-Technologies/com.unity.cinemachine/pull/790

### Testing status

- [ ] Added an automated test
- [x] Manually tested (in our samples scenes)

### Documentation status

### Package version

### DIFF
--- a/com.unity.cinemachine/Runtime/Helpers/InputAxisController.cs
+++ b/com.unity.cinemachine/Runtime/Helpers/InputAxisController.cs
@@ -294,7 +294,7 @@ namespace Cinemachine
         /// <returns>Returns the value of the input device.</returns>
         protected virtual float ReadInput(InputAction action, IInputAxisSource.AxisDescriptor.Hints hint)
         {
-#if true
+#if false
             // GML Temporary fix until this issue is sorted out
             switch (hint)
             {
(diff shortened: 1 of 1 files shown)


com.unity.cinemachine/Runtime/Helpers/InputAxisController.cs
```diff
@@ -294,32 +294,23 @@ static float ReadLegacyInput(string axisName)
...
             
             var actionControlType = activeControl.valueType;
-            if (actionControlType == typeof(float))
-                return action.ReadValue<float>();
-            if (actionControlType == typeof(Vector2))
+            if (actionControlType == typeof(Vector2) || action.expectedControlType == "Vector2")
                 return hint == IInputAxisSource.AxisDescriptor.Hints.Y
                     ? action.ReadValue<Vector2>().y
                     : action.ReadValue<Vector2>().x;
+            if (actionControlType == typeof(float))
```
glabute, line 306: should we also check here for action.expectedControlType == "float"?

glabute (approved):
Please review my changes

gaborkb (commented):
Approve - tested with an integer input which was correctly caught by the catch block.

glabute:
Is integer input a common enough thing that we should support it explicitly?  What about bool?

gaborkb:
I think for our cases no. For our cases Vector2 and float should be the only ones. 
I think it is very easy to extend. If somebody is using some other type and have trouble extending, we can provide them a simple script on the forum.

PR #812, opened 2023-03-01
CMCL-0000: Add parenting warning message

glabute:
### Purpose of this PR

Add message in inspector when CinemachineCamera is parented to CMBrain or another CmCamera

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested


gaborkb (approved):
I think `HelpBoxMessageType.Error` is a bit strong. I can image some cases where Do Nothing cameras can be children of another non-mamager cm cam (maybe on different channels).

But overall, I think this is a usual pitfall for our users, so it is a great add. 👍

PR #815, opened 2023-03-01
CMCL-1392: CinemachineTriggerActionEditor in uitk

glabute:
### Purpose of this PR

What the title says

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
@@ -99,8 +99,8 @@ namespace Cinemachine.Editor
                 if (obj is T)
                     return true;
                 var go = obj as GameObject;
-                if (go == null && obj is Component)
-                    go = (obj as Component).gameObject;
+                if (go == null && obj is Component c)
+                    go = c.gameObject;
                 if (go != null)
                     return go.TryGetComponent<T>(out _);
                 return false;
--- a/com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
@@ -1,152 +1,111 @@
+#if CINEMACHINE_PHYSICS || CINEMACHINE_PHYSICS_2D
 using UnityEngine;
 using UnityEditor;
 using UnityEngine.Playables;
+using UnityEditor.UIElements;
+using UnityEngine.UIElements;
 
 namespace Cinemachine.Editor
 {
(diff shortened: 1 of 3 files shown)


com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
```diff
@@ -1,152 +1,111 @@
...
-                    targetGameObject = targetBehaviour.gameObject;
-                if (targetBehaviour is T)
-                    return targetBehaviour as T;
-                if (targetGameObject != null && targetGameObject.TryGetComponent<T>(out var t))
-                    return t;
+                if (obj is T)
+                    return true;
+                var go = obj as GameObject;
+                if (go == null && obj is Component)
+                    go = (obj as Component).gameObject;
```
gaborkb, line 103: You can cast directly because we know obj is Component here.
```
if (go == null && obj is Component)
     go = ((Component)obj).gameObject;
```
or
```
 if (go == null && obj is Component component)
     go = component.gameObject;
```

gaborkb (changes requested):

com.unity.cinemachine/Editor/Utility/CmPipelineComponentInspectorUtility.cs
```diff
@@ -62,7 +62,7 @@ public enum RequiredTargets { None, Tracking, LookAt, Group };
                             case RequiredTargets.Group: noTarget |= t.FollowTargetAsGroup == null; break;
                         }
                     }
-                    else
+                    else if (targets[i] is CinemachineExtension)
```
gaborkb, line 65: Same here as above.

Can cast x directly:
`var x = (CinemachineExtension)targets[i];`
or
`else if (targets[i] is CinemachineExtension x)`

com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
glabute: > You can cast directly because we know obj is Component here.
I don't understand why that is any better than what is written.  Didn't we discuss this already?

gaborkb: That was slightly different, I think. 

I think it was about using `as` instead of casting directly, but when `is` is used, then the situation is different.

gaborkb (approved):

com.unity.cinemachine/Editor/Editors/CinemachineTriggerActionEditor.cs
glabute: You didn't answer the real question, which was implied by: "I don't understand why that is any better than what is written."

PR #821, opened 2023-03-06
CMCL-0000: Null reference error with StateDrivenCamera.

gaborkb:
### Purpose of this PR
When opening StateDrivenCamera inspector, I got a null reference error thrown from StateDrivenCamera's Editor's `drawElementCallback`.

It was because `m_TargetStateNames` and `m_CameraCandidates` were null.

I had this error in 2023 URP while testing https://github.com/Unity-Technologies/com.unity.cinemachine/pull/819.

### DIFF
--- a/com.unity.cinemachine/Editor/Editors/CinemachineStateDrivenCameraEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineStateDrivenCameraEditor.cs
@@ -61,7 +61,18 @@ namespace Cinemachine.Editor
                 serializedObject.ApplyModifiedProperties();
             });
 
+            ux.TrackAnyUserActivity(() =>
+            {
+                UpdateTargetStates();
+                layerSel.choices = m_LayerNames;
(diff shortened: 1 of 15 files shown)


com.unity.cinemachine/Editor/Editors/CinemachineStateDrivenCameraEditor.cs
```diff
@@ -89,7 +89,20 @@ public override VisualElement CreateInspectorGUI()
             this.AddChildCameras(ux, null);
             this.AddExtensionsDropdown(ux);
 
+            UpdateStatesAndCache();
+            ux.TrackAnyUserActivity(UpdateStatesAndCache);
```
glabute, line 93: This code is already being called, above.  Is this error still reproduceable?

glabute (changes requested):
I think this PR is no longer necessary.  Is the problem reproduceable on main?

gaborkb:
Fixed on main already.

PR #824, opened 2023-03-10
CMCL-0000: input refactor

glabute:
### Purpose of this PR

Input controllers were a bit rigid, difficult to implement third-party input readers.
Refactored to be template-based, allowing the implementor to provide the lowest-level axis reader class.

Adjusted the editor for CinemachineInputAxisController and made it public, so that its logic can be leveraged by editors for other InputAxisController implementations.  Attaching one such implementation (KeyboardInputController.cs), which I used for testing.

[KeyboardInputController.zip](https://github.com/Unity-Technologies/com.unity.cinemachine/files/10945122/KeyboardInputController.zip)

If multiple InputAxisControllers classes are defined, the Add Controller button will display a popup selector:
image

**ALSO:** Added CinemachineDefaultInputActions asset and use that when creating defaults.  Only one player defined for now.  Added DeltaTimeScaleProcessor to normalize mouse movements.

Architecture looks like this:
```
CM Component (e.g. OrbitalFollow)
	- Holds input axes, e.g:
		InputAxis1
			SetValue(float)
			Reset()
		InputAxis2
			SetValue(float)
			Reset()
	- Exposes axes so that they can be discovered dynamically and addressed

IInputReader (interface)
	- provides float value from user

DefaultAxisDriver
	- Helper class to drive InpuAxis.value according to input reader value and settings (accel, decel)

InputAxisController<T> where T: IInputReader:
	- discovers InputAxes
	- uses DefaultAxisDriver to drive the axis value
	- For each discovered axis, exposes driver parameters (accel, decel) and Enabled flag
	- For each discovered axis, exposes InputReader of type T
	- handles reset events from axis owners

Standard Implementations:
------------------------

DefaultInputReader : IInputReader
	- provides float value from Input actions or Legacy input or both (replicates current implementation)

CinemachineInputAxisController : InputAxisController<DefaultInputReader>
	- equivalent to current implementation


Options for controlling axes:
-----------------------------

Level 1: Use CinemachineInputAxisController, then map input actions and/or legacy inputs
Level 2: Use custom InputAxisController with custom IInputReader: take the input from somewhere else (e.g. Rewired, GUI sliders)
Level 3: Whatever you want: call CMComponent.InputAxis1.SetValue directly, with custom algorithm, for custom reasons
```

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### Comments to reviewers

- I did not extract the axis discovery/control management code to an "extension" class, because it didn't really make sense to separate it from its behaviour component: it's all dependent on discovering other behaviours in the same object.
- I was unable to conveniently expose SetValue and Reset methods for each axis so they can be driven by UnityEvents.  Instead, someone who wants to do this will have to write an intermediate class with the methods that the UX buttons can invoke.


AntoineCharton (approved):
Really nice! Makes inputs so much easier to extend and explain.

PR #829, opened 2023-03-15
CMCL-0000: Expose upgrade functionality

gaborkb:
### Purpose of this PR
Virtual Production team would like to access the cm2-cm3 upgrade functionality, because they have the VirtualCameraActor component that uses Cinemachine. They would like to provide the upgrade functionality without losing the reference to the vcam in their component.

They could expose internals like we did for HDRP, but I think this can be useful to other packages/users too.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Package version

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - New IgnoreTarget blend hint will blend rotations without considering the tracking target.
+- Cinemachine 2 to Cinemachine 3 upgrade functionality is now public.
 
 ### Changed
 - All namespaces changed from "Cinemachine" to "Unity.Cinemachine".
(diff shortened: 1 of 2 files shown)


glabute:
I'm uncomfortable with this, because it's not an API we've put much thought into.  Once it's public, we can't take it back.  I think I'd prefer to keep it internal for now.

glabute:
Can you describe what exactly the use-case is, and how they would want to use the upgrader?

gaborkb:
I showed them how to expose internal like we did for HDRP. They are ok with that too, so there is no need to make it public. I understand and share your concern.

Their use-case:
`VirtualCameraActor` references `CinemachineVirtualCamera` when Cinemachine 2 is installed.
`VirtualCameraActor` references `CinemachineCamera` when Cinemachine 3 is installed.

This means if a user upgrades from 2 to 3. Their reference is lost. They cannot use `VirtualCameraBase`, because that would mean a lot of API changes.

They want to add a button when CM3 is installed to `VirtualCameraActor` when they detect that the child gameobject of `VirtualCameraActor` has a `CinemachineVirtualCamera`. The button will upgrade `CinemachineVirtualCamera` and set thee reference  `VirtualCameraActor`. 
Their recommended structure is to have the `CinemachineVirtualCamera` or `CinemachineCamera` as a child of the `CinemachineVirtualCamera`. They don't support other cases.

PR #832, opened 2023-03-15
CMCL-0000: Check if installed Cinemachine is 3 or newer assembly definition

gaborkb:
### Purpose of this PR
LiveCapture team needs this (`CINEMACHINE_3_OR_NEWER`), so they can #if the public API that exposes our internal upgrader code. Similarly how I did it with HDRP, but HDRP luckily had a check that I could use.

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Editor/Unity.Cinemachine.Editor.asmdef
+++ b/com.unity.cinemachine/Editor/Unity.Cinemachine.Editor.asmdef
@@ -83,6 +83,11 @@
             "name": "com.unity.inputsystem",
             "expression": "1.0.0",
             "define": "CINEMACHINE_UNITY_INPUTSYSTEM"
+        },
+        {
+            "name": "com.unity.cinemachine",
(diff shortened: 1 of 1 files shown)


glabute (approved):

glabute:
I don't understand how this will work.  The define is only visible inside the CM assembly

gaborkb:
> I don't understand how this will work. The define is only visible inside the CM assembly

Yes. They create an assembly reference to Cinemachine.Editor. Then they create a script that is going to see our internals, because it is within our assembly. This script now also sees this define.

gaborkb:
@bechrifaAmine will test with this branch first. If all is good on their side, we can merge it.

bechrifaAmine:
> @bechrifaAmine will test with this branch first. If all is good on their side, we can merge it.
It works, thank you

bechrifaAmine (approved):

PR #833, opened 2023-03-16
CMCL-1420 better BlenderSettings inspector

glabute:
### Purpose of this PR

It was unclear that camera specifications in blender settings are strings and do not necessarily have to match any cameras currently in the scene.  The old dropdown hid this.  Now it's more clear.  Also updated the tooltips.

image

Note: you have to undefine USE_IMGUI_INSTRUCTION_LIST in CinemachineBlenderSettingsEditor.cs in order to see the changes, because they're UITK-only.


gaborkb (changes requested):
In 2023.1.0n8 I cannot open the blender dropdowns:
Screen Shot 2023-03-20 at 9 32 16 AM

gaborkb (approved):

PR #837, opened 2023-03-18
CMCL-0000: fix legacy upgrade mechanism and more

glabute:
### Purpose of this PR

Several bugfixes:

- The legacy upgrade mechanism (not to be confused with the upgrader) was broken.  Because it was not called synchronously, there was no guarantee that it got called after the relevant fields we properly initialized.  Moved it to where it's guaranteed to be called properly.
- ForceCameraPostion() was overflowing the stack - fix was to not call it for the parent (parent is calling it for the children - duh)
- vcam.GetInputAxisProvider() was re-added for use by deprecated classes

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Runtime/Deprecated/CinemachineInputProvider.cs
+++ b/com.unity.cinemachine/Runtime/Deprecated/CinemachineInputProvider.cs
@@ -1,7 +1,8 @@
-﻿#if CINEMACHINE_UNITY_INPUTSYSTEM
-using System;
-using System.Linq;
+﻿using System;
 using UnityEngine;
+
+#if CINEMACHINE_UNITY_INPUTSYSTEM
+using System.Linq;
 using UnityEngine.InputSystem;
 using UnityEngine.InputSystem.Users;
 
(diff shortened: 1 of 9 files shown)


com.unity.cinemachine/Runtime/Deprecated/CinemachineInputProvider.cs
```diff
@@ -132,3 +132,25 @@ namespace Unity.Cinemachine
     public class CinemachineInputProvider : MonoBehaviour {}
 }
 #endif
+
+namespace Unity.Cinemachine
+{
+    [Obsolete("IInputAxisProvider is deprecated.  Use InputAxis and InputAxisController instead")]
```
gaborkb, line 138: `using System` is missing for this. It is hidden by the #if

gaborkb (approved):
Missing using statement.

com.unity.cinemachine/Runtime/Deprecated/CinemachineInputProvider.cs
glabute: Good catch, thanks for checking!

PR #843, opened 2023-03-23
CMCL-1414: Runtime UI TK for Brain Debug message

gaborkb:
### Purpose of this PR
[CMCL-1414](https://jira.unity3d.com/browse/CMCL-1414): Replace Debug Text implementation with UITK

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Comments to reviewers
1. I hooked into the existing OnGUIHandler pipeline. We should refactor this and probably make CinemachineDebug handle it.

2. To create the Runtime UI, I am adding a global gameobject to the scene. This gameobject stores a UI Document, which is needed to get a panel to the scene with which we can position correctly (without any hacks) our visual elements. It is hidden from the hierarchy, it won't be part of any build. It is managed by CinemachineDebug.

3. New look reference:
debugui


com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
```diff
@@ -674,5 +711,71 @@ void PushStateToUnityCamera(ref CameraState state)
             // Send the camera updated event
             CinemachineCore.CameraUpdatedEvent.Invoke(this);
         }
+
+#if UNITY_EDITOR
+        class DebugText : IDisposable
```
glabute, line 716: I'm feeling like the brain is having to do too much work.  I think I'd prefer these details to be hidden inside CinemachineDebug.  Can you think of a way to do that?

I'm imagining CinemachineDebug to be a service that assigns text widgets to any clients (e.g. CmBrain) who register and request a place to print on-screen.  Positioning and layout is 100% managed by CinemachineDebug.  Client can pass in a Camera object when registering.  All client has to do is update the text, color, and visibility status of the widget.  Implementation details are hidden.

com.unity.cinemachine/Runtime/Core/CinemachineDebug.cs
```diff
@@ -31,6 +35,62 @@ public static Rect GetScreenPos(Camera camera, string text, GUIStyle style)
...
+                        tag = "EditorOnly"
+                    };
+                }
+            }
+            if (!s_UIDocumentHolder.TryGetComponent(out s_UIDocument))
+            {
+                s_UIDocument = s_UIDocumentHolder.AddComponent<UIDocument>();
+                
+                const string path = "Packages/com.unity.cinemachine/Runtime/UI/";
+                s_UIDocument.panelSettings = AssetDatabase.LoadAssetAtPath<PanelSettings>(path + "CinemachinePanelSettings.asset");
```
glabute, line 62: This line is the only thing that's preventing the CinemachineDebug service to be usable at runtime - which could be an interesting thing to have.  Can you think of a way to build this asset in-memory?

glabute (changes requested):

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,138 @@
...
+                        hideFlags = HideFlags.NotEditable | HideFlags.HideInHierarchy,
+                        tag = "EditorOnly" // TODO: we may want runtime too
+                    };
+                }
+            }
+            if (!s_UIDocumentHolder.TryGetComponent(out s_UIDocument))
+            {
+                s_UIDocument = s_UIDocumentHolder.AddComponent<UIDocument>();
+                
+                const string path = "Packages/com.unity.cinemachine/Runtime/UI/";
```
glabute, line 63: Should this asset be in the Debug directory?

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,138 @@
...
+                name = "CinemachineDebugUI_ViewportContainer",
+                style =
+                {
+                    position = new StyleEnum<Position>(Position.Absolute),
+                    flexWrap = new StyleEnum<Wrap>(Wrap.Wrap),
+                }
+            };
+            // need to use delayCall, because rootVisualElement may not be built at this point yet
+            EditorApplication.delayCall += () => s_UIDocument.rootVisualElement.Add(viewportContainer);
+            viewportContainer.schedule.Execute(() => PositionWithinCameraView(viewportContainer, outputCamera)).Every(0); // TODO: can we do better?
```
glabute, line 81: This is going to be running whether or not there is debug text displayed.  I'm sure we can improve that. We should be updating the containers only if they're actually doing any work.

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,138 @@
...
+            s_CameraViewRectContainers.Add(outputCamera, viewportContainer);
+            
+            return viewportContainer;
+            
+            static void PositionWithinCameraView(VisualElement visualElement, Camera camera)
```
glabute, line 86: How can the user control the size of the text?  There needs to be a way.  Failing that, at the very minimum text should be the same size as editor text, to ensure legibility.  I don't understand how the size is determined now, but on my machine it's not legible.

image

com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
```diff
@@ -197,8 +204,11 @@ void OnEnable()
             m_BlendManager.OnEnable();
 
             s_ActiveBrains.Add(this);
+#if UNITY_EDITOR
+            m_DebugText = new DebugText(OutputCamera);
```
glabute, line 208: This should only be created if needed.  Creating it has a performance impact because it creates and updates the panels every frame.

glabute (changes requested):

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
gaborkb: Yes!

com.unity.cinemachine/Runtime/Behaviours/CinemachineBrain.cs
glabute: Debug service should cost 0 unless it's actively being used

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
gaborkb: Fixed with panel settings' scaling option.

com.unity.cinemachine/Runtime/Core/CinemachineDebug.cs
gaborkb: I'll make a separate task for making it work in build too. I talked to UITK team a bit and it is not a clear.

"It is really preferable you create it in the editor as we are using the panel settings to include some assets in builds, like the theme and the shaders"
So we would need to create a persistent debug ui gameobject that's in all scenes hidden in case user wants to debug in build.

They said there may be a generic build debug console in the future ("not soon") that we could use instead.

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,149 @@
...
+        public static VisualElement CreateRuntimeUIContainer(Camera outputCamera)
+        {
+            if (s_UIDocumentHolder == null)
+            {
+                s_UIDocumentHolder = GameObject.Find(k_DebugUIName);
+                if (s_UIDocumentHolder == null)
+                {
+                    s_UIDocumentHolder = new GameObject(k_DebugUIName)
+                    {
+                        hideFlags = HideFlags.NotEditable | HideFlags.HideInHierarchy,
```
glabute, line 54: Missing DontSave flag - very important

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,149 @@
...
+
+#if CINEMACHINE_UNITY_IMGUI
+        /// <summary>Reserve an on-screen rectangle for debugging output.</summary>
+        /// <param name="client">The client caller.  This is used as a handle.</param>
+        /// <param name="text">Sample text, for determining rectangle size</param>
+        /// <param name="style">What style will be used to draw, used here for
+        /// determining rect size</param>
+        /// <returns>An area on the game screen large enough to print the text
+        /// in the style indicated</returns>
+        public static Rect GetScreenPos(Camera camera, string text, GUIStyle style)
```
glabute, line 25: This is no longer used - remove the whole section and the CINEMACHINE_UNITY_IMGUI define from the asmdef

com.unity.cinemachine/Runtime/Unity.Cinemachine.asmdef
```diff
@@ -76,6 +76,11 @@
             "name": "com.unity.modules.imgui",
             "expression": "1.0.0",
             "define": "CINEMACHINE_UNITY_IMGUI"
```
glabute, line 78: Remove CINEMACHINE_UNITY_IMGUI define, not needed anymore.  Do it in the editor asmdef also

glabute (changes requested):

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
gaborkb: Why?

gaborkb: UITK team said that it is preferable to have the UI document set in editor, because it needs to set some shaders before playing.

I tried with DontSave option, but then the result is inconsistent. Sometimes, we get UI, sometimes we don't.

glabute: DontSave will prevent the object from being saved with the scene.  There is no reason to save it with the scene - it's a volatile thing that we should create it on an as-needed basis.  We do this elsewhere in CM.  Look in the code for `HideFlags.HideAndDontSave`

> UITK team said that it is preferable to have the UI document set in editor, because it needs to set some shaders before playing.

I don't see the relevance of that

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,126 @@
...
+{
+    /// <summary>Manages onscreen positions for Cinemachine debugging output</summary>
+    static class CinemachineDebug
+    {
+        static List<StringBuilder> s_AvailableStringBuilders;
+        
+#if UNITY_EDITOR && CINEMACHINE_UIELEMENTS
+        const string k_DebugUIName = "CinemachineDebugUI";
+        static Dictionary<Camera, VisualElement> s_CameraViewRectContainers = new();
+        static GameObject s_UIDocumentHolder;
```
glabute, line 20: This stuff is getting added to the scene.  When you open a new scene, what happens to it?  You need to rethink this.  The idea is that this should be a volatile object, not saved with the scene, and created as-needed.  It will be lost when the scene is unloaded.

com.unity.cinemachine/Runtime/Debug/DebugText.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            public void SetTextColor(Color color) => m_DebugLabel.style.color = new StyleColor(color);
+            public void RestoreOriginalTextColor() => m_DebugLabel.style.color = m_OriginalTextColor;
+            public void SetText(string text) => m_DebugLabel.text = text;
+
+            public void Dispose()
+            {
+                ReleaseUnmanagedResources();
+                GC.SuppressFinalize(this);
```
glabute, line 47: Can you explain this?  I've never seen it before.

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
```diff
@@ -0,0 +1,149 @@
...
+                    RuntimeUtility.DestroyObject(s_UIDocumentHolder);
+                    s_UIDocumentHolder = null;
+                }
+
+                if (s_UIDocumentHolder == null)
+                {
+                    s_UIDocumentHolder = new GameObject(k_DebugUIName)
+                    {
+                        hideFlags = HideFlags.HideAndDontSave,
+                        tag = "EditorOnly"
```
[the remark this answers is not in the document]
gaborkb: The gameObject will be removed from build.

glabute (approved):
Looks good.  Just remove the EditorOnly tag thing if it's not needed.  Pushed some cosmetic changes.

com.unity.cinemachine/Runtime/Debug/CinemachineDebug.cs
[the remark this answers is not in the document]
glabute: So it's a built-in magic tag?  I see.  In fact it's redundant because UNITY_EDITOR is not defined in builds, but it doesn't hurt, I guess.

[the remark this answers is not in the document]
gaborkb: Yes, it is a built-in magic tag. And yes, it is redundant. But I thought it is cleaner this way.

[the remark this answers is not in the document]
glabute: ok, keep it in

PR #847, opened 2023-03-28
CMCL-0000: Upgrade samples to uitk, reorganize

glabute:
### Purpose of this PR

- Use new UITK-based UX in all the samples.
- Delete old OnGUI-based scripts
- Removed all scene folders, all assets moved to SharedAssets.
- Removed 2D-local assets, everything now in common SharedAssets
- Fixed broken input asset controllers to work with Input package
- Fixed broken blend hints due to refactor

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Comments to reviewers

I noticed several issues, will make a new Jira ticket for them


com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
```diff
@@ -184,7 +184,7 @@ static bool PromptUserImportDependencyConfirmation(List<string> dependencies)
             }
         }
 
-        static readonly string[] k_MaterialFolders = { "Materials", "Cameron/Model" };
+        static readonly string[] k_MaterialFolders = { "Materials" };
```
gaborkb, line 187: we could get rid of this and the for loop below.

gaborkb: FadeOut material is not convertible, so we'll have to make an exception for that, because it is now mixed with all other materials.

gaborkb (changes requested):
Clearshot's folder was not deleted:
Screen Shot 2023-03-28 at 1 49 21 PM

------------------
Wrong API call in CinemachinePixelPerfect (URP):
`/Runtime/Behaviours/CinemachinePixelPerfect.cs(35,41): error CS1061: 'CinemachineBrain' does not contain a definition for 'IsLive' and no accessible extension method 'IsLive' accepting a first argument of type 'CinemachineBrain' could be found (are you missing a using directive or an assembly reference?)`

------------------

com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs
glabute: ClearShot may have been a metafile issue, since the folder and scene have the same name.  Pushed a different metafile.

glabute: URP API issue is not related to this PR, but will fix here

PR #864, opened 2023-04-18
CMCL-1443: Solo mode should play the sequencer

glabute:
### Purpose of this PR

Sequencer was not playing when in Solo mode

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Editor/Utility/CmCameraInspectorUtility.cs
+++ b/com.unity.cinemachine/Editor/Utility/CmCameraInspectorUtility.cs
@@ -118,10 +118,7 @@ namespace Unity.Cinemachine.Editor
 
                     // Refresh the game view if solo and not playing
                     if (isSolo && !Application.isPlaying)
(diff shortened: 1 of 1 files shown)


AntoineCharton:
Just wandering what if a user script relies on the state of the sequencer? What if someone want to see a state but in the middle of a sequence? Can we keep the solo button just changing which camera is focused and have an other button called solo play (Or a setting to disable it)?

glabute:
> Just wandering what if a user script relies on the state of the sequencer

AntoineCharton:
> > Just wandering what if a user script relies on the state of the sequencer
> 
> Not sure I understand this question.

As a user do you expect a button called "Solo" to do anything else than soloing the camera?

glabute:
> As a user do you expect a button called "Solo" to do anything else than soloing the camera?

uh, no

gaborkb (approved):

PR #865, opened 2023-04-20
CMCL-0000: 2D platformer fixes

gaborkb:
### Purpose of this PR
A couple of minor fixes:
- Warning caused by overriding parent classes CameraOverride property. -> Fix: remove in child and use parent's.
- Removed duplicate capsule collider from Player 2D prefab.

Found bug: Left camera is never activated, because it depends on the Follow Target's rotation and not on the PlayerGeometry (Animated Cameron's rotation).
Possible fix is to add a `Transform PlayerGeometry` property to `PlatformerCamera2D` like `SimplePlayerController2D`, and using that to determine left/right orientation.

### Testing status

- [ ] Added an automated test
- [x] Manually tested (in URP and built-in)

### Documentation status

### Package version


glabute (approved):
Pushed some changes:
- There are supposed to be 2 capsules on Player2D.
- We don't want CameraOverride  in the base class.
- Good catch on LeftCamera, added a fix

glabute:
Also added missing namespaces

gaborkb:
It looks good now

PR #869, opened 2023-04-24
CMCL-0000: All samples reference shared input map

AntoineCharton:
Make all samples use shared input map instead of the default in cinemachine package

### Purpose of this PR

More consistency in the samples 

### Technical risk

Low samples


glabute (approved):
Pushed some tweaks

PR #870, opened 2023-04-29
CMCL-1473: Improve handling of nested blends

glabute:
### Purpose of this PR

CMCL-1473: Improve handling of nested blends.

1. Blends ((A -> B) -> A) -> B have incorrect timing (can zip fast at end).  This should not happen.  Instead, blends should take the appropriate remaining blend time.
2. Nested blends should continue evaluating, so that (A -> B) -> C will move the camera in an arc instead of taking a segmented path.
3. Blends should not nest indefinitely, so that ((A -> B) -> C) -> D should take a snapshot of A -> B

See the attached scene in the jira ticket for a good test case.  Cycle between the 3 cameras to test blend chaining.

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### Note to reviewers: 

This PR is included in https://github.com/Unity-Technologies/com.unity.cinemachine/pull/871, which is built off of this branch

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [unreleased]
+## [Unreleased]
 
 ### Fixed
 
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [unreleased]
+
+### Changed
+- Improve handling of nested blends
+
+
 ## [3.0.0-pre.5] - 2023-04-25
 
 ### Fixed
--- a/com.unity.cinemachine/Runtime/Core/BlendManager.cs
+++ b/com.unity.cinemachine/Runtime/Core/BlendManager.cs
@@ -8,7 +8,7 @@ namespace Unity.Cinemachine
     /// It takes care of looking up BlendDefinitions when blends need to be created,
     /// and it generates ActivationEvents for camweras when necessary.
     /// </summary>
-    internal class BlendManager : CameraBlendStack
+    class BlendManager : CameraBlendStack
     {
         // Current Brain State - result of all frames.  Blend camB is "current" camera always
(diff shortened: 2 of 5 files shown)


gaborkb (commented):
Debug label no longer specifies if we are blending from a blend.

Image below after blending from vcam 1 to vcam 2 about half way and then starting a blend back to vcam 1.
Screen Shot 2023-05-02 at 12 38 04 PM


EDIT: that was not the case before either. But it specified the previous blends out direction in the from part. So the above message was CM vcam 1 from CM vcam 2
Screen Shot 2023-05-02 at 12 38 04 PM

gaborkb (approved):

AntoineCharton (approved):
Nice!

Screen Shot 2023-05-02 at 1 29 38 PM

Wandering what should happen with cut and a blend time set to 0 when nested with a blend of 5second?

gaborkb (changes requested):
I don't think this behaviour is correct:

https://user-images.githubusercontent.com/57672095/235743122-6446fdc9-e74a-42d2-83db-81168650c327.mov

In the recording, I click to blend from vcam1 to 3, then to 2 (at around 20%), then to 1 (at around 20%). What's weird is that the blend overshots (it should not go above the blue box and then back)

1 -> 3
(1->3 at 20%) -> 2
((1->3 at 20%) to 2 at 20%) -> 1

gaborkb:
> I don't think this behaviour is correct:
> 
>  Screen.Recording.2023-05-02.at.1.38.14.PM.mov 
> In the recording, I click to blend from vcam1 to 3, then to 2 (at around 20%), then to 1 (at around 20%). What's weird is that the blend overshots (it should not go above the blue box and then back)
> 
> 1 -> 3 (1->3 at 20%) -> 2 ((1->3 at 20%) to 2 at 20%) -> 1
> 
> Cam1 -> looks at blue Cam2 -> looks at yellow Cam3 -> looks at red

I talked to Antoine and it makes sense that it does not fit all cases perfectly. But the new custom blend (https://github.com/Unity-Technologies/com.unity.cinemachine/pull/872) fixes this issue by enabling the user to fit the blending behaviour to their need.

glabute:
The blending is correct, I think.

The new nested blending behaviour continues to evaluate the blends, even as new blends are added on to of them.  So this blend (red -> yellow) -> blue, and red -> yellow continues to evaluate, pushing the start point above the blue box.  To change that, you have to set this BlendHint in the camera, which will cause a snapshot to be taken of the nested blend. 
image

gaborkb (approved):
I understand. Makes sense :)

glabute:
> Debug label no longer specifies if we are blending from a blend.
> 
> Image below after blending from vcam 1 to vcam 2 about half way and then starting a blend back to vcam 1. Screen Shot 2023-05-02 at 12 38 04 PM
> 
> EDIT: that was not the case before either. But it specified the previous blends out direction in the from part. So the above message was CM vcam 1 from CM vcam 2: Screen Shot 2023-05-02 at 12 43 24 PM

Yes, I will fix.  Good catch

PR #875, opened 2023-05-02
CMCL-0000: Fix evaluate local orientation returning global value

AntoineCharton:
Fix of evaluate local orientation returning a global value instead of a local one. 

Reported by a user.
https://forum.unity.com/threads/dolly-track-evaluate-orientation-is-incorrect-if-track-is-rotated-cm-2-9-5.1432078/#post-8987869

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineSmoothPath.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineSmoothPath.cs
@@ -221,11 +221,9 @@ namespace Cinemachine
                         m_ControlPoints2[indexA].roll, m_Waypoints[indexB].roll);
                 }
 
-                Vector3 fwd = EvaluateTangent(pos);
+                Vector3 fwd = EvaluateLocalTangent(pos);
                 if (!fwd.AlmostZero())
                     result = Quaternion.LookRotation(fwd) * RollAroundForward(roll);
-                
-                result = Quaternion.Inverse(transform.rotation) * result;
             }
             return result;
         }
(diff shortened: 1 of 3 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineSmoothPath.cs
```diff
@@ -224,6 +224,8 @@ public override Quaternion EvaluateLocalOrientation(float pos)
                 Vector3 fwd = EvaluateTangent(pos);
                 if (!fwd.AlmostZero())
                     result = Quaternion.LookRotation(fwd) * RollAroundForward(roll);
+                
+                result = Quaternion.Inverse(transform.rotation) * result;
```
glabute: Yes, this was a regression introduced within the last year or so.
However, the fix is not optimal.  Pushing a better fix.

glabute (approved):
Please check my changes

com.unity.cinemachine/Runtime/Behaviours/CinemachineSmoothPath.cs
```diff
@@ -221,7 +221,7 @@ public override Quaternion EvaluateLocalOrientation(float pos)
                         m_ControlPoints2[indexA].roll, m_Waypoints[indexB].roll);
                 }
 
-                Vector3 fwd = EvaluateTangent(pos);
+                Vector3 fwd = EvaluateLocalTangent(pos);
```
AntoineCharton, line 224: Yes better than what I had

PR #905, opened 2023-10-02
CMCL-0000: Fix tests

AntoineCharton:

### Purpose of this PR

Fix tests


com.unity.cinemachine/ValidationExceptions.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ErrorExceptions": [
+    {
+      "ValidationTest": "Unity Version Validation",
+      "ExceptionMessage": "In package.json, \"unity\" is pointing to a version higher (2022.3) than the editor you are currently using (2022.2.21f1). Validation needs to happen on a version of the editor that is supported by the package.",
```
glabute, line 5: I don't understand this fix.  The correct thing is to stop testing against 2022.2.
Why is the unity version 2022.2.21f1 hardcoded in the message?

glabute (changes requested):

com.unity.cinemachine/ValidationExceptions.json
AntoineCharton: Testing it

glabute (approved):

PR #912, opened 2023-11-24
CMCL-0000:add some missing offsets

glabute:
### Purpose of this PR

HardLookAtTarget was missing a target offset - it always looked at target's root.
GroupFraming was missing a CenterOffset - it always centered the group on the screen.

Added these missing items.


AntoineCharton:
Nov-27-2023 14-06-46
Why isn't it offsetting in this configuration?

AntoineCharton (approved):
Just check the typo but the features works has expected.

PR #916, opened 2023-11-28
CMCL-0000: Update image version to fix job failing on MacOS

AntoineCharton:
package-ci/stable has been deprecated for all platforms. Updated images versions to fix the issue.


glabute:
I have no idea what this is.
Everything is failing, is that normal?

PR #924, opened 2024-01-23
CMCL-0000: Samples upgrade for improved versatility

glabute:
### Purpose of this PR

The sample scenes and scripts are a useful starting point for beginners to Unity.  This PR increases the number of usecases that they cover by making them work better on arbitrary surfaces:
- SimplePlayerAimController.cs upgraded to work on arbitrary surfaces (no longer depends on world up).  It's now possible, for instance to use it on a character that can walk on a sphere.
- SimplePlayerController.cs has logic to avoid input gimbal lock when player is upside-down relative to the camera.  This allows for situations where the player can walk upside-down while the camera stays upright.
- PlayerOnSphere.cs is now PlayerOnSurface - can walk on arbitrary surfaces, does not assume a sphere.  The code is substantially more complex, and uses raycasts to discover the surface.

Also added comments and tooltips where they were lacking.


AntoineCharton (approved):
Works for me! 
Fixed also the split screen sample not having the input action reference.

PR #925, opened 2024-01-25
CMCL-0000: Add sample editor to prevent copy paste mistakes

AntoineCharton:
### Purpose of this PR

- Limit the risk of forgetting to copy paste a file
- Limit the risk of breaking samples that use Shared assets

### Testing status

Internal tool.

### Technical risk

Internal tool.

### Comments to reviewers

To edit a sample:
- Open the SamplesEditor project.
- Load the samples
Screenshot 2024-01-25 at 3 48 40 PM
- Edit your changes
- Then hit save to copy them to the samples folder of the package


glabute (changes requested):
This is a good idea, but there is an important flaw.  There are some assets that get "fixed up" if project is HDRP or uses new input system.  The default versions of these assets must be compatible with builtin and legacy input - so those assets must NOT get copied back to the repo.  See here for code that does the asset fixup: https://github.com/Unity-Technologies/com.unity.cinemachine/blob/59b0fd48c32247c7a7c9e6a6c7ca47eecfb30811/com.unity.cinemachine/Editor/Samples/SampleDependencyImporter.cs#L214

AntoineCharton:
Closing until i get the time to look at it. Tryed but there is no shortcut to it. Will need to think about it maybe refactor the way we import samples.

PR #927, opened 2024-01-30
CMCL-0000: Better filtering for vcam Navel Gazing warning

glabute:
### Purpose of this PR

With this vcam setup, there should not be a "camera is looking at itself" warning, because the camera doesn't try to look at the LookAt target:

image

This PR adds an internal API so the inspector can ask a component if it's trying to look at the LookAt target, and display the warning only if it is.

### Testing status

- [ ] Added an automated test
- [x] Manually tested


AntoineCharton (approved):
Nice catch! Will check the fly around sample with this fix.

PR #928, opened 2024-02-04
Changing `CinemachineSplineDolly.PositionUnits` now adjusts the `CinemachineSplineDolly.CameraPosition` value to stay at the same world position.

Walter-Hulsebos:
### Purpose of this PR

Currently, changing a `CinemachineSplineDolly`'s `PositionUnits` does not change the value of `CameraPosition`, except for some clamping. This PR changes that.

**Before:**
https://github.com/Unity-Technologies/com.unity.cinemachine/assets/77513543/1c03e09d-17ab-4b29-ae2f-1a7792c4f022
_Adjusting the `PositionUnits` doesn't adjust the value to stay in the same relative position._
(Notice how the camera jumps around when you change it?)

**After:**
https://github.com/Unity-Technologies/com.unity.cinemachine/assets/77513543/5c79ed41-23ba-4fab-9fec-8bfac76f6ed8
_Adjusting the `PositionUnits` DOES adjust the value to stay in the same relative position._
(Notice how the cameras stays at the same position when you change it?)

This enhancement also applies when adjusting `PositionUnits` via code, not just the inspector.

### Testing status

I have implemented PlayMode unit tests to validate the correct modification of the `CameraPosition` when changing the `PositionUnits`. 
Additionally, I've included tests to ensure that values remain unchanged when setting the `PositionUnits` to the same value. Manual testing has been performed on the inspector functionality, and the CM2 to CM3 upgrade process has been validated.

- [x] Added an automated test
- [x] Manually tested 

### Documentation status

Not sure what to modify in user documentation here, but I did add summaries to all the public/internal code I've added.

### Technical risk

This PR introduces a change in the behaviour of modifying `PositionUnits`, which now modifies`CameraPosition` as well. 
Users with existing scripts relying on the previous behaviour will encounter a different outcome. 
Overall, I think that tradeoff is worth it for the extra ease of use.
If users want to keep the same value after setting PositionUnits, they can save the position before changing the units and reapply it after calling. 

Perhaps an additional method could be added to modify `PositionUnits` *without* modifying `CameraPosition`, not unlike the `.SetValueWithoutNotify` that exists for UI Elements? Let me know what you think.

### Comments to reviewers

There's one major thing I'm not sure about, and that's the `ArgumentOutOfRangeException`s I throw in `SplineContainerExtensions.cs`, perhaps it would be better to loop around the input for them if it's incorrect? 
We have the spline data so it's trivial to do, on the other hand, it's fault input and they could just do such a loop around themselves before manually calling `ConvertDistance` and friends, which are internal anyway.
Cinemachine's internal code doesn't call it with faulty input.

### Package version

- [ ] Updated package version

### DIFF
--- a/com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
@@ -1,6 +1,7 @@
 using UnityEditor;
 using UnityEditor.UIElements;
 using UnityEngine.UIElements;
+using UnityEngine.Splines;
 
 namespace Unity.Cinemachine.Editor
 {
@@ -22,9 +23,11 @@ namespace Unity.Cinemachine.Editor
 
             var row = ux.AddChild(InspectorUtility.PropertyRow(
                 serializedObject.FindProperty(() => Target.CameraPosition), out _));
-            row.Contents.Add(new PropertyField(serializedObject.FindProperty(() => Target.positionUnitsBackingField), "") 
-                { style = { flexGrow = 2, flexBasis = 0 }});
-
+            
+            SerializedProperty positionUnitsProp  = serializedObject.FindProperty(() => Target.positionUnitsBackingField);
+            PropertyField      positionUnitsField = new (positionUnitsProp, label: "") { style = { flexGrow = 2, flexBasis = 0 } };
+            row.Contents.Add(positionUnitsField);
+            
             ux.Add(new PropertyField(serializedObject.FindProperty(() => Target.SplineOffset)));
             ux.Add(new PropertyField(serializedObject.FindProperty(() => Target.CameraRotation)));
             ux.Add(new PropertyField(serializedObject.FindProperty(() => Target.AutomaticDolly)));
@@ -38,6 +41,18 @@ namespace Unity.Cinemachine.Editor
                 for (int i = 0; !noSpline && i < targets.Length; ++i)
                     noSpline = targets[i] != null && ((CinemachineSplineDolly)targets[i]).Spline == null;
                 noSplineHelp.SetVisible(noSpline);
+            }
+            
+            // Store the initial value
+            PathIndexUnit previousUnits = Target.positionUnitsBackingField;
+            ux.TrackPropertyValue(positionUnitsProp, TrackPositionUnits);
+            void TrackPositionUnits(SerializedProperty p)
+            {
+                PathIndexUnit newUnits = (PathIndexUnit) p.enumValueIndex;
+                
+                Target.UpdateDistanceForPositionUnits(previousUnits, newUnits);
+                // Update previous value for the next change
+                previousUnits = newUnits;
             }
 
             return ux;
--- a/com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
+++ b/com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
@@ -22,7 +22,7 @@ namespace Unity.Cinemachine.Editor
 
             var row = ux.AddChild(InspectorUtility.PropertyRow(
                 serializedObject.FindProperty(() => Target.CameraPosition), out _));
-            row.Contents.Add(new PropertyField(serializedObject.FindProperty(() => Target.PositionUnits), "") 
+            row.Contents.Add(new PropertyField(serializedObject.FindProperty(() => Target.positionUnitsBackingField), "") 
                 { style = { flexGrow = 2, flexBasis = 0 }});
 
             ux.Add(new PropertyField(serializedObject.FindProperty(() => Target.SplineOffset)));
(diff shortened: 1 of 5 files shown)


com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
```diff
@@ -39,6 +42,18 @@ void TrackSpline(SerializedProperty p)
...
             }
+            
+            // Store the initial value
+            PathIndexUnit previousUnits = Target.positionUnitsBackingField;
+            ux.TrackPropertyValue(positionUnitsProp, TrackPositionUnits);
+            void TrackPositionUnits(SerializedProperty p)
+            {
+                PathIndexUnit newUnits = (PathIndexUnit) p.enumValueIndex;
+                
+                Target.UpdateDistanceForPositionUnits(previousUnits, newUnits);
```
glabute, line 53: This code is problematic because it doesn't handle undo or multi-object selection

com.unity.cinemachine/Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -33,18 +33,41 @@ public class CinemachineSplineDolly : CinemachineComponentBase
...
+        
         [Tooltip("How to interpret the Spline Position:\n"
             + "- Distance: Values range from 0 (start of Spline) to Length of the Spline (end of Spline).\n"
             + "- Normalized: Values range from 0 (start of Spline) to 1 (end of Spline).\n"
             + "- Knot: Values are defined by knot indices and a fractional value representing the normalized " 
             + "interpolation between the specific knot index and the next knot.\n")]
-        public PathIndexUnit PositionUnits = PathIndexUnit.Normalized;
+        [SerializeField] 
+        [FormerlySerializedAs(oldName: "PositionUnits")]
+        internal PathIndexUnit positionUnitsBackingField = PathIndexUnit.Normalized;
```
glabute, line 44: This is a little messy in my opinion.  I do like that it preserves serialization back-compatibility.

com.unity.cinemachine/Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -131,6 +154,7 @@ public struct DampingSettings
 
         CinemachineSplineRoll m_RollCache; // don't use this directly - use SplineRoll
 
+        #if UNITY_EDITOR
```
glabute, line 157: This needs to be runtime also

com.unity.cinemachine/Runtime/Core/SplineContainerExtensions.cs
```diff
@@ -138,5 +145,190 @@ static class SplineContainerExtensions
...
+        /// <summary>
+        /// Converts the distance from one unit to another.
+        /// </summary>
+        /// <param name="spline">The spline which length is used for conversion.</param>
+        /// <param name="distance">The distance to convert.</param>
+        /// <param name="oldUnits">The original units of the distance.</param>
+        /// <param name="newUnits">The units to convert the distance to.</param>
+        /// <returns> The converted distance.</returns>
+        [MethodImpl(methodImplOptions: AggressiveInlining)]
+        internal static float ConvertDistance(this Spline spline, float distance, PathIndexUnit oldUnits, PathIndexUnit newUnits)
```
glabute, line 158: This re-implements functionality that is already available in SplineUtility.ConvertIndexUnit

com.unity.cinemachine/Tests/Runtime/SplineDollyCameraTest.cs
```diff
@@ -153,6 +153,127 @@ public IEnumerator LinearInterpolationTestPathUnits()
             yield return null;
             UnityEngine.Assertions.Assert.AreApproximatelyEqual(Vector3.Distance(m_CmCam.State.GetFinalPosition(), new Vector3(7, 1, -2.5f)), 0, 0.1f);
         }
+
+        [UnityTest]
+        public IEnumerator PositionUnits_ChangesCameraPositionCorrectly_WhenPositionUnitsChanged()
```
glabute, line 158: Very good of you to implement these tests :) Thank you

com.unity.cinemachine/Editor/Editors/CinemachineSplineDollyEditor.cs
Walter-Hulsebos: Good point, bad oversight of mine!

com.unity.cinemachine/Runtime/Core/SplineContainerExtensions.cs
Walter-Hulsebos: I overlooked this functionality entirely. Thanks!

com.unity.cinemachine/Runtime/Components/CinemachineSplineDolly.cs
Walter-Hulsebos: Your refactor indeed offers a much more elegant solution. 
Initially, I figured converting it to use a Property would be fairly frictionless for end-users but I agree it's messy.

PR #929, opened 2024-02-06
CMCL-0000: Improvements to PR-928: Spline position units

glabute:
### Purpose of this PR
I really like the initiative of [PR-928](https://github.com/Unity-Technologies/com.unity.cinemachine/pull/928), it significantly improves the UX and is worth implementing.  However, I do have problems with the implementation:
- No implementation for SplineCart
- Incorrect handling of Undo and multi-select in inspector

I've refactored it here by introducing a SplinePosition struct with its own PropertyDrawer.  This allows it to be easily used by multiple clients, including SplineDolly and SplineCart, and potentially other custom classes provided by the user.

The downside of this PR is that it breaks backwards-compatibility for serialization of SplineCart and SplineDolly (because members have moved into a new struct and there is no simple way to handle such changes).  I'm willing to bite the bullet on that one, and just document the issue in the changelog.

Samples were updated for this change.

### Testing status

- [x] Added an automated test
- [x] Manually tested 

### Documentation status


com.unity.cinemachine/Runtime/Behaviours/CinemachineSplineCart.cs
```diff
@@ -33,29 +34,39 @@ public enum UpdateMethods
         [Tooltip("When to move the cart, if Speed is non-zero")]
         public UpdateMethods UpdateMethod = UpdateMethods.Update;
 
-        /// <summary>How to interpret the Spline Position</summary>
-        [Tooltip("How to interpret the Spline Position.  If set to Knot, values are as follows: "
-            + "0 represents the first knot on the path, 1 is the second, and so on.  "
-            + "Values in-between are points on the spline in between the knots.  "
-            + "If set to Distance, then Spline Position represents distance along the spline.")]
-        public PathIndexUnit PositionUnits = PathIndexUnit.Distance;
```
AntoineCharton, line 41: We can't change the API without a major bump 4.X.X

AntoineCharton (changes requested):
Change is awesome! We would need a major bump for the changes to the API. Can we make the new fields internal and revert the naming changes?

glabute:
Maybe we can make an exception.  IMO this isn't enough of a reason to bump the major version.

com.unity.cinemachine/Runtime/Behaviours/CinemachineSplineCart.cs
glabute: Changed it to break less API

glabute:
I changed it so that less API is broken (new properties have the same names as the old fields).  Maybe that is good enough?

Walter-Hulsebos:
> I changed it so that less API is broken (new properties have the same names as the old fields). Maybe that is good enough?

It would still make it so prior configurations are lost, since this wouldn't preserve serialized values from before this change, no?

glabute:
> It would still make it so prior configurations are lost, since this wouldn't preserve serialized values from before this change, no?
Yes, unfortunately.  With a little effort I could possibly fix that (stream into some private place using FormerlySerlializedAs, then find an opportune moment to patch the values into the right place).  There are places in CM where this sort of thing happens, but it's always a little iffy imo (search for PerformLegacyUpgrade).

com.unity.cinemachine/Runtime/Components/CinemachineSplineDolly.cs
```diff
@@ -21,30 +21,10 @@ namespace Unity.Cinemachine
...
-        /// - Distance: Values range from 0 (start of Spline) to Length of the Spline (end of Spline).
-        /// - Normalized: Values range from 0 (start of Spline) to 1 (end of Spline).
-        /// - Knot: Values are defined by knot indices and a fractional value representing the normalized
-        /// interpolation between the specific knot index and the next knot."</summary>
-        [Tooltip("How to interpret the Spline Position:\n"
-            + "- Distance: Values range from 0 (start of Spline) to Length of the Spline (end of Spline).\n"
-            + "- Normalized: Values range from 0 (start of Spline) to 1 (end of Spline).\n"
-            + "- Knot: Values are defined by knot indices and a fractional value representing the normalized " 
-            + "interpolation between the specific knot index and the next knot.\n")]
-        public PathIndexUnit PositionUnits = PathIndexUnit.Normalized;
```
AntoineCharton, line 47: What happen's with serialization? It breaks the running sample scene.

glabute: Are you checking with freshly-installed samples?  I'm supposed to have upgraded the running race scene

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.cinemachine?pullRequest=929) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.cinemachine?pullRequest=929) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: Walter-Hulsebos<br/>:x: glabute<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.cinemachine?pullRequest=929) it.</sub>

AntoineCharton (approved):
Works but for 3.1.X only do not backport plz.

PR #939, opened 2024-02-23
CMCL-0000: add isDelayed to some fields

glabute:
### Purpose of this PR

Typing values in certain text fields can be confusing because validation kicks in as you're typing and changes the value unintuitively.  Solution is to wait for Enter or focus loss before validating.

Added this property to the UX of some fields:
- Lens FOV
- Composition ranges (PositionComposer, RotationComposer)
- Axis Value/Center/Range

If later we find more fields that need this, we can fix in a similar manner.

### Testing status

- [ ] Added an automated test
- [x] Manually tested


AntoineCharton (approved):
That's such a great idea!!!

PR #951, opened 2024-03-15
CMCL-1391: storyboard editor in uitk

glabute:
### Purpose of this PR

Part of the IMGUI purge epic.
Should be no visible difference in the CinemachineStoryboard editor or the Waveform Monitor window.


AntoineCharton (commented):
Same as
https://github.com/Unity-Technologies/com.unity.cinemachine/pull/952

PR #952, opened 2024-03-15
CMCL-1396: postprocessing and volumesettings editors in uitk

glabute:
### Purpose of this PR

Ongoing purge of IMGUI.
PostProcessing and VolumeSettings inspectors should look the same as before.


AntoineCharton (commented):
Is this expected?
Mar-26-2024 13-16-34

AntoineCharton (commented):
What am I doing wrong?
Screenshot 2024-03-26 at 1 19 50 PM

glabute:
> Is the float on the position field expected?
Yes, position is a float field.

> What am I doing wrong?
Probably you have to enable the child properties of Focus Mode - but I think there is a problem here with the UX not updating properly.  Will look.
Behaviour of the inspector should be identical to before this commit.

> Can we postpone this one? It will take at least a good day to review properly.
Yes.

PR #955, opened 2024-03-18
CMCL-1576:  Spline roll calculated incorrectly when spline rotated

glabute:
### Purpose of this PR

**This fix should be backported to 3.0 branch.**

https://forum.unity.com/threads/3-0-1-spline-cart-rotates-wrongly-if-the-splinecontainer-is-rotated.1559951/

A spline cart that's following a rotated spline container winds up with an incorrect rotation

rotation = rotation * spline.transform.rotation;
This applies the spline container's rotation after the evaluated rotation, which is backwards. This results in weird orientations that wobble around as you rotate the spline container.

It should go like this:

rotation = spline.transform.rotation * rotation;
After the fix, the cart behaves as expected. This code is in SplineContainerExtension.cs in version 3.0.1.

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -7,10 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


AntoineCharton (approved):
Nice it's possible there was a Jira ticket for that. 

For later maybe consider adding a unit test for this use case.

PR #964, opened 2024-03-29
Properly reset Snapshots when blending.

### LINKED ISSUE #963: The Blend Hint "Freeze When Blending Out" doesn't workj when switching betwen two cameras multiple times
When switching between Cinemachine Cameras, one of them having the Blend Hinter "Freeze When Blending Out", the freeze only works the first time I switch to the camera. The second time the position seems to be set to 0,0,0

The issue seems to be caused by not resetting m_SnapshotSource in GetSnapshotIfAppropriate when the Snapshot is reset.

`if (cam == null || (cam.State.BlendHint & CameraState.BlendHints.FreezeWhenBlendingOut) == 0)
{
    // No snapshot required - reset it
    m_Snapshot.TakeSnapshot(null);
    return cam;
}
// A snapshot is needed
if (m_SnapshotSource != cam || m_SnapshotBlendWeight > weight)
{
    // At this point we're pretty sure this is a new blend,
    // so we take a new snapshot of the camera state
    m_Snapshot.TakeSnapshot(cam);
    m_SnapshotSource = cam;
    m_SnapshotBlendWeight = weight;
}`
  
The first time I blend to the camera m_SnapshotSource is null (weights are 0). 
Then I switch back to the first camera, m_Snapshot gets reset.

Second blend back to the camera with Freeze, no new Snapshot is taken as m_SnapshotSource  is still the camera taken during the first blend, but the actual snapshot has already been reset (0,0,0 position etc.)
The obcious solution seems to be to set to m_SnapshotSource  to null when the snapshot is reset.

SvenRH:

### Purpose of this PR

Fixes #963 
When using the Blend Hint "Freeze When Blending Out" the blend snapshot was not properly reset causing issues when blending to the same camera again (using an empty snapshot, causing the camera to jump to 0,0,0).

### Testing status

- [ ] Added an automated test
- [X ] Manually tested

### DIFF
--- a/com.unity.cinemachine/Runtime/Core/CameraBlendStack.cs
+++ b/com.unity.cinemachine/Runtime/Core/CameraBlendStack.cs
@@ -82,6 +82,8 @@ namespace Unity.Cinemachine
                 {
                     // No snapshot required - reset it
                     m_Snapshot.TakeSnapshot(null);
+                    m_SnapshotSource = null;
+                    m_SnapshotBlendWeight = 0;
                     return cam;
                 }
                 // A snapshot is needed


glabute (approved):
Awesome! I don't know how we missed this.  Thanks.

PR #966, opened 2024-04-04
CMCL-1579: add SaveDuringPlay confirmation dialog

glabute:
### Purpose of this PR

CMCL-1579: Add confirmation when exiting play mode is SaveDuringPlay is enabled and changes have been made.

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested


AntoineCharton:
Apr-04-2024 15-56-14
Is it supposed to pop it even if no changes were made?

glabute:
>Is it supposed to pop it even if no changes were made?

Technically no.  Let me try to see why that's happening

glabute:
Hah!  It was a stupid implementation.  Good catch.

I redid it, and now it always applies the changes, but if you select "cancel" in the dialog it just does an Undo for you.  So in fact you can still access the changes in that case by hitting Redo.

AntoineCharton (approved):
Awesome!

PR #967, opened 2024-04-05
CMCL-1580: scene always dirtied 2.x

glabute:
### Purpose of this PR

CMCL-1580: Cinemachine dirties scenes for nothing.  This is a long-standing issue.

Fix is to set the vcam and main camera transform's position and rotation only if significantly different from what it is.  There is one gotcha: because transform rotation is stored in local space, setting it in global space will introduce imprecisions when conversion is made.  So we do the delta check in local space, and set the values in local spacee also.

Because all that takes processing power, we do it only if in editor and not playing.

**NOTE: This needs to be done for 3.X also.**

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low


AntoineCharton:
Screenshot 2024-04-09 at 2 16 46 PM
PR has compilation issues with 2019 and 2020. Can we fix those?

AntoineCharton (approved):

PR #972, opened 2024-04-11
CMCL-0000: Retool the helpbox with button UX according to guidelines

glabute:
### Purpose of this PR

The UX design team has decided how helpbox with buttons should look.  See https://jira.unity3d.com/browse/UXQA-469.

This PR moves our implementation closer to the style guidelines (for IMGUI it's not exact because the infrastructure is not there, but we got it as close as we can).

IMGUI:
image

UITK:
image

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Note to reviewers:

InspectorUtility.cs has some code moved around, but otherwise unchanged.  The relevant diffs are in the two HelpBoxWithButton() methods.

### Technical risk

low


AntoineCharton (approved):
Ok

PR #973, opened 2024-04-11
CMCL-0000: remove some isdelayed attributes because it interferes with dynamic refresh

glabute:
### Purpose of this PR

A recent PR had added isDelayed to some float fields (e.g. FOV, axis value) because when typing values OnValidate() got called and interfered with the typing.  However, the side-effect of this was that when dragging the label to change the value, no dynamic refresh happened until dragging stopped.

We think that the cure is worse than the disease, so we are reverting the isDelayed while we look for a better solution.

### Testing status

- [ ] Added an automated test
- [x] Manually tested


AntoineCharton (approved):
Would be interesting to revisit this in the future but yeah definitely better disabled in the current state.

PR #974, opened 2024-04-12
CMCL-0000: Add isDelayed to some fields in a drag-tolerant way

glabute:
### Purpose of this PR

The IsDelayed function on float fields delays processing of the text input until enter is pressed.  This is useful because otherwise OnValidate() kicks in and interferes with your typing when it does its clamping.

However, isDelayed also prevents processing when you drag on the label, until you stop dragging.  This kills the visual feedback that is crucial for some fields.

This PR re-introduces dragging for isDelayed fields by installing a custom dragger that temporarily disables isDelayed while the property is being dragged.

This custom dragger has been added to the following fields:

- FOV/Ortho size (collapsed and expanded versions)
- Axis Value (collapsed and expanded versions)
- Composition settings in PositionComposer and RotationComposer.  

### Testing status

- [ ] Added an automated test
- [x] Manually tested


AntoineCharton (approved):
Thanks for fixing this!

PR #975, opened 2024-04-18
CMCL-0000: Group Framing works with Confiner 2D

glabute:
### Purpose of this PR

GroupFraming and Confiner2D did not play well together.  
Slack thread: https://unity.slack.com/archives/C4P4KJR9A/p1713367131830879

This PR introduces a special compatibility mode to handle this common use-case.  GroupFraming checks for the presence of Confiner2D and if it finds one, it runs in the Body stage (with an early execution order) so that it happens before Confiner2D.  Also, in this mode it monitors the ortho size and invalidates the confiner lens cache when the ortho size changes.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
@@ -133,11 +133,17 @@ namespace Unity.Cinemachine
             public Vector2 RotAdjustment;
             public float FovAdjustment;
 
+            public CinemachineCore.Stage Stage = CinemachineCore.Stage.Finalize; // uninitialized state
+            public CinemachineConfiner2D Confiner;
+            public float PreviousOrthoSize;
+
             public void Reset()
             {
                 PosAdjustment = Vector3.zero;
                 RotAdjustment = Vector2.zero;
                 FovAdjustment = 0;
+                Stage = CinemachineCore.Stage.Finalize;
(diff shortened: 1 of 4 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineGroupFraming.cs
```diff
@@ -133,11 +133,17 @@ class VcamExtraState : VcamExtraStateBase
             public Vector2 RotAdjustment;
             public float FovAdjustment;
 
+            public CinemachineCore.Stage Stage = CinemachineCore.Stage.Finalize; // uninitialized state
+            public CinemachineConfiner2D Confiner;
+            public float PreviousOrthoSize;
```
AntoineCharton, line 138:
```suggestion
            [SerializedField]
            private float m_PreviousOrthoSize;
```
This is adding API. If we do that we need a minor bump not a patch. Can we make it private or internal instead? Maybe we could start planning for 3.2?

glabute: It's not new API, it's an internal private class

AntoineCharton: Oh ok yeah mb

AntoineCharton (approved):
Nice change!

PR #981, opened 2024-05-05
CMCL-1583: fix camera update logic, brain update modes sample

glabute:
### Purpose of this PR

Brain Update Modes sample was broken, indicating deeper problems with the camera update logic.

**Problem:**
The SmartUpdate window is not smart enough: red car is jittery when it has focus.  It gets smooth if you disable the Fixed Update World.  Something was wrong with the target tracking: multiple brains interfere with each other somehow.

**Solution:**
- Fixed Camera Update logic to better support multiple CM brains in heterogeneous update modes
- Cm Cameras must only update once per frame: either Render frame or Physics frame, depending on the update mode of the brain.  If Smart Update, then update time for each camera will depend on the update method of the camera's target.
- Camera targets are now updated correctly in the target tracker (once per physics frame, AND once per render frame).
- BrainUpdateModes sample was tweaked for clarity (help text, camera names)

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low to medium: need to make sure that all cameras are properly updated in various circumstances


AntoineCharton (approved):
Under which condition was the jittering happening? Seems to work but I wasn't able to reproduce with 3.1.0.

glabute:
Just running the scene, in the smart update window the currently-tracked car should be smooth and the other one jittery (it alternates between the two cars with a sequencer)

PR #989, opened 2024-05-09
CMCL-1587: cleanup UX layout code. fix alignment for U6

glabute:
### Purpose of this PR

Fix a number of UX issues:

- A number of CM inspectors were not vertically aligned
- Helpbox with button was messed up on U6
- FreeLookModifier item foldout arrows were not visible
- Moved some obsolete code in InspectorUtility to the obsolete folder (needed for CM2 support only)
- Refactored the InspectorUtility layout helpers to support tabbing navigation
- Hid the Legacy Targets member of TargetGroup
- Refactored all the list views to no longer use `bindItem`, which changed behaviour as of 2023.3.31f1.  This includes
  - SequencerCamera
  - StateDrivenCamera
  - CustomBlendSettings (in Brain and manager camera inspectors)
  - SplineRoll
  - SplineLookAtTargets
  - ChildCamera lists in all manager cameras

### Testing status

- [ ] Added an automated test
- [x] Manually tested on U6 and 2022.3


com.unity.cinemachine/Editor/Obsolete/ObsoletePropertyDrawers.cs
```diff
@@ -158,7 +158,7 @@ public override void OnGUI(Rect rect, SerializedProperty property, GUIContent la
         }
     }
 
-    partial class InputAxisWithNamePropertyDrawer : PropertyDrawer
+    partial class InputAxisPropertyDrawer : PropertyDrawer
```
AntoineCharton, line 161: Why do we need to rename an obsolete class?

glabute:
because it's partial and connects to a non-obsolete renamed class (which is internal in any case)

AntoineCharton (approved):
Not sure why we needed some changes on some internal deprecated classes? 
Any thoughts on setting weight parameters on mixingCamera has NoSaveDuringPlay? It is prompting save on MixingCamera sample.
Other than that looks good tested on U6 and 2022.

PR #993, opened 2024-05-10
CMCL-1589: No SaveDuringPlay for CMBrain

glabute:
### Purpose of this PR

SaveDuringPlay is meant for camera tweaking, not Brain tweaking.  Some users change things like Default Blend at runtime.,  There is no reason to save such changes, and doing so can mess people up.


AntoineCharton (approved):
Great idea.

PR #994, opened 2024-05-10
CMCL-1590: statedrivencamera inspector bugs

glabute:
### Purpose of this PR

1. StateDrivenCamera inspector was failing to correctly set states in the instruction list.
2. Added multi-select support for StateDrivenCamera inspector
3. Also: if you drag a child cam outside of the SDC then press Undo, SDC instructions that were referencing the camera lose the reference (reference does not get restored).
4. Same thing for SequencerCamera.

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

low


AntoineCharton:
Tested on U6 and 2022.3:
- ThirdPersonWithRoadieRun sample
- Switching states
- Adding new state
- Save During play mode
- Change wait time
- Ctrl+Z

Any other idea?

glabute:
That should do it

AntoineCharton (approved):
Only weird behaviour is not forgetting the cinemachine camera if un-parented but even there even with prefabs and nested prefabs it doesn't break. I would keep it this way unless we find a bug around this. Thoughts?

PR #1005, opened 2024-06-04
CMCL-0000: Convert shader to shader graph to support built in, URP and HDRP

AntoineCharton:
### Purpose of this PR

Use shader graph to have cross rp compatibility on fade out shader.
Screenshot 2024-06-05 at 12 31 55 PM

### Test
Manually tested on all 3 render pipeline.

### Comment
- This adds a dependancy to shader graph but it is installed by default on most templates.


glabute (commented):
Is there a reason why the FadeOut material (and the shadergraph) are not in the shared assets folder?

AntoineCharton:
> Is there a reason why the FadeOut material (and the shadergraph) are not in the shared assets folder?
- It is only used on 3D samples and allows us to limit shader graph dependancy to 3d samples only instead of having all samples depending on it.

PR #1018, opened 2024-07-30
CMCL-1603: Added Padding option to CinemachineConfiner2D

glabute:
### Purpose of this PR

CMCL-1603:
https://discussions.unity.com/t/how-to-have-a-2d-cinemachine-track-go-off-in-multiple-directions/372767

When the camera ortho size is the same as the bounding shape, confiner2D allows a small amount of movement when it shouldn't.  On first glance, it seems to be related to the step size used when baking.  A small imprecision later gets magnified by the scaling.

The issue is that an arbitrary padding is added to the minimal confiner shape, to reduce the chance of creating a series of dots instead of a continuous shape.  This PR introduces an option to control the size of this padding.  The user should set it to the smallest value that produces acceptable results.

image

### Testing status

- [ ] Added an automated test
- [x] Manually tested 

### Documentation status

### Technical risk

Low.  Larger padding values reproduce the old behaviour.


eakesy (approved):
Tested for regressions using the Cinemachine 2D Samples project. No issues found with the introduction of this new option.

PR #1045, opened 2025-03-12
CMCL-1654: freelook improvements

glabute:
### Purpose of this PR

CMCL-1654: When the FreeLookModifier modifies the screen composition, the RotationComposer applies damping to the change, producing a sluggish effect:

broken

The desired behaviour is to bypass damping for this change, giving an undamped direct response.  The current behaviour is a regression compared to CM 2.X, which behaved as desired.

This PR addresses this problem, and others:
- RotationComposer damping is bypassed for changes to the ScreenPosition parameter, whether via inspector or FreeLookModifier
- RotationComposer code was optimized for performance.  The UpdateCache method was improved so that it does less work.
- When a FreeLookModifier is actively modifying the composition, the GameView composer guides now show the modified composition instead of misleadingly showing the center value.
- When a FreeLookModifier is actively modifying the composition, the GameView composer guides are no longer draggable since they are displaying the blended result.  Top/center/bottom values have to be set via the inspector.
- FreeLookModifier now supports SaveDuringPlay for tweaks to existing item properties.

The fixed version (there is still a small amount of damping, but it's acceptable):

fixed

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### Documentation status

### Technical risk

low

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -10,10 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Bugfixes
 - Deoccluder did not always properly reset its state.
 - Mac only: Extensions dropdown in CinemachineCamera inspector did not work consistently.
-- Regresion fix: Confiner2D was not always confining when a camera was newly activated.
+- Regression fix: Confiner2D was not always confining when a camera was newly activated.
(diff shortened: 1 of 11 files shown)


eakesy (approved):
Tested changes in sample scene, behaviour is now as expected, what I see now makes sense compared to before, works great from what I can tell... also looked at the other samples to make sure nothing weird happened somewhere else.

sebastienduverne (approved):
I made suggestions, please review them to see if they still correspond to your initial intention. Thanks!

PR #1046, opened 2025-03-14
CMCL-1657: CinemachineDeoccluder damping interferes with FreeLook

glabute:
### Purpose of this PR

CMCL-1657:
This fixes an artifact introduced by a recent change to the Deoccluder damping algorithm.

Open the FreeLook Deoccluder sample scene.
Put a large Damping value on the deoccluder (e.g.3)
Make the top orbit very high, so that the camera is far away compared to the other orbits
Play the scene.
Move the character cloase enough to a wall while looking horizontally to make the camera deocclude.
Rotate on X axis to eliminate the occlusion (camera begins to damp back to position)
Now, go to top orbit.  We expect to get there quickly (because it's a user-triggered action) but damping makes it happen slowly.

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### DIFF
--- a/com.unity.cinemachine/Runtime/Behaviours/CinemachineDeoccluder.cs
+++ b/com.unity.cinemachine/Runtime/Behaviours/CinemachineDeoccluder.cs
@@ -468,7 +468,14 @@ namespace Unity.Cinemachine
                                 var newOffset = initialCamPos + displacement - resolutionTargetPoint;
                                 var newOffsetMag = newOffset.magnitude;
                                 var newOffsetDir = newOffset / newOffsetMag;
+
+                                // Avoid introducing spurious damping when the camera changed position relative to the target.
+                                // We calculate the previous offset from target in two ways, and take the one that's closest
+                                // to the current desired offset.
                                 var prevOffsetMag = extra.PreviousCameraOffset.magnitude;
+                                var prevOffsetMag2 = (initialCamPos - resolutionTargetPoint).magnitude - Mathf.Sqrt(prevDispMag);
+                                if (Mathf.Abs(newOffsetMag - prevOffsetMag2) < Mathf.Abs(newOffsetMag - prevOffsetMag))
+                                    prevOffsetMag = prevOffsetMag2;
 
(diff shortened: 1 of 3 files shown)


com.unity.cinemachine/Runtime/Behaviours/CinemachineDeoccluder.cs
```diff
@@ -468,7 +468,14 @@ protected override void PostPipelineStageCallback(
...
                                 var newOffsetMag = newOffset.magnitude;
                                 var newOffsetDir = newOffset / newOffsetMag;
+
+                                // Avoid introducing spurious damping when the camera changed position relative to the target.
+                                // We calculate the previous offset from target in two ways, and take the one that's closest
+                                // to the current desired offset.
                                 var prevOffsetMag = extra.PreviousCameraOffset.magnitude;
+                                var prevOffsetMag2 = (initialCamPos - resolutionTargetPoint).magnitude - Mathf.Sqrt(prevDispMag);
+                                if (Mathf.Abs(newOffsetMag - prevOffsetMag2) < Mathf.Abs(newOffsetMag - prevOffsetMag))
+                                    prevOffsetMag = prevOffsetMag2;
```
arnaudparevogt-unity, line 478: Should this also be used when the strategy is PreserveCameraHeight?

glabute: Probably yes, but that's a more risky change

arnaudparevogt-unity: makes sense

arnaudparevogt-unity (approved):

com.unity.cinemachine/Runtime/Behaviours/CinemachineDeoccluder.cs
glabute: In the end the fix for the other modes was really easy and risk-free, so I went for it.  Thanks for flagging it!

PR #1053, opened 2025-04-01
CMCL-1665: confiner 2D dynamic float-to-int scaling

glabute:
### Purpose of this PR

CMCL-1665: CinemachineConfiner2D stops working when bounding shape is very large

Conversion between float world coords and the fixed point that Clipper expects was suffering when bounding shapes approaceh 1000 units in size.  This is because the hardcoded scaling factor was tuned for small bounding shapes and was too precise for large shapes.

The fix was to dynamically adapt the scaling factor, keeping it the same for small shapes, but decreasing it gradually as the shapes got bigger.

### Testing status

- [ ] Added an automated test
- [x] Manually tested

### Documentation status

### Technical risk

low


alexzaranek (approved):
Validated that the Confiner now handles larger bounding boxes correctly. Changes looks good.

PR #1107, opened 2026-02-27
Fix for UUM-131977

arnaudparevogt-unity:
### Purpose of this PR

[Jira](https://jira.unity3d.com/browse/UUM-131977) [Issue Tracker](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-131977)

When cancelling a blend, `CameraBlendStack` was using a snapshot of the current state to keep track of the midpoint of the blend. This meant that if one of the two cameras were moving, the midpoint would become outdated.

This commit fixes the issue by introducing FrozenBlendSource an (internal) kind of nested blend source that will not advance the time (AdvanceBlend) but still update the child sources.

### Testing status

- [X] Added an automated test
- [X] Passed all automated tests
- [X] Manually tested

Only manually tested user repro project

### Documentation status

- [X] Updated [CHANGELOG](https://keepachangelog.com/en/1.0.0/)

### Technical risk

Tech: Low    Halo: Low

### Comments to reviewers

I created a subclass of `NestedBlendSource` instead of modifying it directly because it allows scoping the concept of freezing the time of a blend source to the `CameraBlendStack` type.

I refactored a bit `AdvanceBlend` to make it less indented, only functional change should be `CameraBlendStack.cs:345` (not ticking the time if source if FrozenBlendSource).

### Package version

- [ ] Updated package version

### DIFF
--- a/com.unity.cinemachine/CHANGELOG.md
+++ b/com.unity.cinemachine/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.1.7-pre.2] - 2026-03-26
+
+### Bugfixes
+- Fixed blend cancellation not taking into account updates to "cancelled" camera state
+
 ## [3.1.7-pre.1] - 2026-02-24
 
 ### Bugfixes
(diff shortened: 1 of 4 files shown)


glabute (approved):
Very good fix

com.unity.cinemachine/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.1.7-pre.2] - 2026-03-26
+
+### Bugfixes
+- Fixed blend cancellation not taking into account updates to "cancelled" camera state
+
 ## [3.1.7-pre.1] - 2026-02-24
```
sim-bz: Can you replace the 3.1.7-pre.1 header with 3.1.7-pre.2 instead?

We'll just update the changelog with the final version when we do a release and skip the pre release versions. This way we will avoid making a full release with no changes in the changelog.

sim-bz (approved):

